PR #2, opened 2020-11-24
Feat/ci

xgroleau:
What was done:
- Added clang-tidy on compilation for more check
- Added clang-format via `make format` and `make check-format`. 
- Added LGTM
- Added Github CI

Note that the run-clang-format was not written by me. [See link](https://github.com/Sarcasm/run-clang-format)

What was not added
- Sanitizer for thread and memory debugging, arm-none does not support it

Notes: A lot of flags are enable, we will probably remove some of them as development is going.

Sources:
http://www.stablecoder.ca/2018/10/30/full-cmake-helper-suite.html
https://github.com/Sarcasm/run-clang-format


cmake/clang-tools/clang-tidy.cmake
```diff
@@ -0,0 +1,7 @@
+find_program (CLANG_TIDY_EXE NAMES "clang-tidy")
+if(CLANG_TIDY_EXE AND ENABLE_CLANG_TIDY_CHECK)
+    # set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_EXE}" CACHE STRING "" FORCE)
```
cquesnel, line 3: Is this still needed?

cquesnel (approved):

cmake/clang-tools/clang-tidy.cmake
xgroleau: Good catch

PR #3, opened 2020-11-29
added stm32-cmake license

xgroleau:
Added the proper license at the top of  files from external repo

### DIFF
--- a/THIRD-PARTY-NOTICES.md
+++ b/THIRD-PARTY-NOTICES.md
@@ -1,0 +1,61 @@
+# Third party notices
+
+Hivemind uses components from other projects listed below.
+If a required notice is missing, please send an email at swarmus@usherbrooke.ca
+
+
+## run-clang-format
+
+Project host at: https://github.com/Sarcasm/run-clang-format
+
+### License
+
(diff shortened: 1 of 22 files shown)


stlaperriere (approved):

stlaperriere:
I suppose it is customary to write the license in every CMake file? Only writing them in the repository readme is not enough?

xgroleau:
> 
> @stlaperriere 
> I suppose it is customary to write the license in every CMake file? Only writing them in the repository readme is not enough?
 
Since the code was copy-pasted from another repository, we need to provide their license because of this clause of the license
 
**The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.**

I'm not sure if a reference in the readme is enough. I think that with this formulation we need to provide the whole license and not only the link to the repo. We could provide it in another file and specify to which files it applies but I'm not sure if it's better.
See [this link](https://softwareengineering.stackexchange.com/questions/234437/bsd-mit-license-changes-when-copy-pasting-a-function-from-another-bsd-mit-licens). I'll check if we could regroup them in a folder for another PR

PR #4, opened 2020-12-02
Feat/add doxygen support

xgroleau:
Added doxygen to build.

Note that the documentation is build on `make all` and throws error on warnings so no need to check for it in the CI.

We can add a theme eventually if necessary since the default looks pretty outdated.

Sources:
https://cmake.org/cmake/help/latest/module/FindDoxygen.html


.github/workflows/ci.yml
```diff
@@ -13,6 +13,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: prepare
       run: |
+        sudo apt-get install -y doxygen
```
g-lesssard, line 16: Also needs graphviz to generate png images. We should also add clang-tidy.

g-lesssard (changes requested):

g-lesssard:
All good! Worked for me after installing the required dependencies. I flagged one missing dependency on the set-up for the ci action. It does not fail the build but the images (class diagrams and such) were missing.

cquesnel:
I think we should add the required packages in the .readme as @g-lesssard has stated, we should at a minimum include doxygen, graphviz and clang-tidy as those are less likely to be already installed on peoples systems

cmake/doc/doxygen.cmake
```diff
@@ -0,0 +1,40 @@
...
+    set(DOXYGEN_MAX_DOT_GRAPH_DEPTH 0)
+    set(DOXYGEN_DOT_TRANSPARENT YES)
+
+    if(ENABLE_WARNINGS_AS_ERROR )
+        set(DOXYGEN_WARN_AS_ERROR YES)
+    endif()
+
+    doxygen_add_docs(doc
+        ALL
+        ${PROJECT_SOURCE_DIR}/src
```
cquesnel, line 35: Do you think we should maybe remove the tests from the documentation? I have a feeling it will end up cluttering the documentation.

xgroleau: I'll add them. Should cmake hard fail if doxygen and clang-tidy is missing or simply throw an error message? Obviously I would add a flag to disable the hard fail if needed, but it would be on by default.

.github/workflows/ci.yml
xgroleau: Good catch!

cmake/doc/doxygen.cmake
cquesnel: I'm guessing you're commenting on the .readme suggestion?
If so, I think just a warning as we currently have is enough, it's not like if we absolutely need to generate the doc or check clang-tidy when building

g-lesssard (approved):

cquesnel (approved):

PR #5, opened 2020-12-12
Feat/logging

xgroleau:
What was done

- Added Logging capabilities
- Added Print capabilities
- Added support for stmcubemx
- Fix warnings as error on poject, was just affecting main.cpp
- Added syscalls to smt32

Sorry, the branch went a bit out of scope and became way bigger than anticipated


src/bsp/include/bsp/ui_impl.h
```diff
@@ -0,0 +1,14 @@
+#ifndef __UI_IMPL_H_
+#define __UI_IMPL_H_
+
+#include "bsp/ui.h"
+
+class UIImpl : public UI {
```
cquesnel, line 6: Are we naming the concrete class with a `Impl` suffix?
Wouldn't it be better to add a I to the interface instead?

src/bsp/src/stm32f4/hal/stm32f429zi/Core/Inc/hivemind_hal.h
```diff
@@ -0,0 +1,21 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "gpio.h"
+#include "main.h"
+#include "stm32f4xx_hal_conf.h"
+#include "usart.h"
+
+#define huart_print (&huart3)
```
cquesnel, line 13: I can't find it in the LLVM coding standard, but by experience, shouldn't constants be all caps?

src/bsp/src/stm32f4/hal/stm32f429zi/Core/Src/main.c
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * @brief  Retargets the C library printf function to the USART.
+ * @param  None
+ * @retval None
+ */
+PUTCHAR_PROTOTYPE {
+    /* Place your implementation of fputc here */
+    /* e.g. write a character to the USART3 and Loop until the end of transmission */
+    HAL_UART_Transmit(&huart3, (uint8_t*)&ch, 1, HAL_MAX_DELAY);
```
cquesnel, line 74: Not using the define from `hivemin_hal.h`
And I think we should transmit using a non-blocking function with either interrupt of DMA

src/logger/include/logger/logger_impl.h
```diff
@@ -0,0 +1,26 @@
...
+#include "logger/logger.h"
+#include <bsp/ui.h>
+
+#include <FreeRTOS.h>
+#include <semphr.h>
+
+/*
+ * @brief TODO: change if namespace
+ */
+class LoggerImpl : public Logger {
```
cquesnel, line 13: Same comment as before on the use of `Impl`

src/logger/src/logger_impl.cpp
```diff
@@ -0,0 +1,40 @@
+#include "logger/logger_impl.h"
+#include <bsp/ui.h>
+#include <cstdarg>
+
+LoggerImpl::LoggerImpl(LogLevel level, UI* ui) {
+    m_log_level = level;
+    m_ui = ui;
```
cquesnel, line 7: We should define our standard on null checks if we decide to do them, there is one missing here

cquesnel: We could maybe use class references instead of pointers when possible
https://isocpp.org/wiki/faq/references

cquesnel (changes requested):

src/bsp/include/bsp/ui_impl.h
xgroleau: Yea I thought about this, what does the I prefix actually brings? The user of the UI library doesn't need to know that it's an interface, only the maintainer needs to know that. We need to add interface to everything anyway since we have unit testing and the mocking requires it. The I prefix is only a standard in C#, for exemple in Java the List, Collection, Iteratable are interfaces but are not prefixed with I.

I knew this would come up in the PR. @stlaperriere @t-constantin @g-lesssard what do you guys think? Let's just pick a standard and go on with it, as long as it's consistent. The I prefix is fine if the other agrees, I'm just not sure what benefits it brings.

See this post for a bit more info:
 https://softwareengineering.stackexchange.com/questions/117348/should-interface-names-begin-with-an-i-prefix

src/freertos/platforms/posix/include/freertos_platform_config.h
```diff
@@ -0,0 +1,8 @@
+#ifndef __FREERTOS_PLATFORM_CONFIG_H_
+#define __FREERTOS_PLATFORM_CONFIG_H_
+
+/* Clock config that is platform dependent */
+
```
g-lesssard, line 5: By standards, shouldn't defines be all caps, in this case, CONFIG_TICK_RATE_HZ?

src/logger/include/logger/logger.h
```diff
@@ -0,0 +1,37 @@
...
+#define __LOGGER_H_
+
+/**
+ * @brief The log level used for the logger;
+ */
+enum class LogLevel { Debug = 0, Info = 1, Warn = 2, Error = 3 };
+
+/**
+ * @brief The return value of the logger
+ */
```
g-lesssard, line 11: What does "LowLevel" stands for? Is it an error?

src/logger/src/logger_impl.cpp
g-lesssard: In this case, the UI class should be a singleton. We should manage it using an std:shared_pointer. This way, we don't have to do a null check on the parameter passed since it would be caught on the creation of the shared_pointer. Having a shared pointer would also add protection for multithread use. It has relatively low overhead for embedded systems (https://stackoverflow.com/questions/22295665/how-much-is-the-overhead-of-smart-pointers-compared-to-normal-pointers-in-c).

src/logger/include/logger/logger.h
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @param [in] format Text to be written, can contain format specifiers that will be replace by
+     *values specified in the additionnal arguments, matches the standard printf function
+     *
+     * @param [in] ... Additionnal arguments for the format parameter
+     *
+     * @return Returns Ok on success, LowLevel if the level provided is lower than the current one
+     *and Error if an error occured
+     */
+    virtual LogRet log(LogLevel level, const char* format, ...) = 0;
```
g-lesssard, line 34: I would add a method to change the log level at run time. Might be useful in the future.

src/bsp/include/bsp/ui_impl.h
cquesnel: I don't mind not using `I`, but if not using it means we have `Impl` on concrete classes I'd rather have the `I`

g-lesssard (changes requested):

src/freertos/platforms/posix/include/freertos_platform_config.h
cquesnel: This is a FreeRTOS config, it's not done by us

src/bsp/include/bsp/ui_impl.h
g-lesssard: The main reason is to differentiate from the class implementation with some sort of consistency. Either we add something to the implementation of the interface definition. The same interface could be implemented by multiple classes (IState for example) and is, therefore, more efficient. Also, a class does not require to implement an interface, so a trailing Impl would be useless, and not having it might be confusing to differentiate it from interfaces. 

For the part of the user knowing that the class is an interface, why is it a problem if the user knows? On the contrary, I would advise that it should be known if it needs to be mocked for unit testing on his side. By design, the UI class exposed to the user should be an interface, so having a prefix to the class only confirms it for the user.

TLDR: a prefix or suffix should be added to differentiate an interface from its implementation and it seems more efficient to add it to the interface. It would also add clarity for external users, which might not be needed but wouldn't be harmful either.

stlaperriere: My many decades of writing oo code in Java (sike) have never brought me to having to name my interfaces the same way as I name the implementations. So I never really had this problem. But then again, in Java the interfaces are specifically declared as such, and not as classes that have their virtual methods `=0`. And I've never really done much unit-testing in an object-oriented language.

Anyway. I don't like the idea of having to prefix the interfaces with I, nor do I like the idea of having to write `impl` or anything of the sort at the end. I feel like an interface should be different enough from an implementation that there should not even be any naming conflicts (to me, an interface simply describes a few elements of behaviour and should not describe the whole of a class; what would even be the point of the interface if it is _that_ specific?). 

But I gather that in order to write some unit-tests, we need to "force" the use of interfaces. In this case, I would choose the lesser evil and go with the prefix I, and lose the `impl`.

src/bsp/include/bsp/ui.h
```diff
@@ -0,0 +1,29 @@
+#ifndef __UI_H_
+#define __UI_H_
+
+/**
+ * @brief Manages the user interface
+ * The user interface can consist of buttons, LED, serial print ports, etc.
+ * It is not a graphical interface.
+ * The board will communicate it's state via the UI and the user can send commands via the UI.
```
stlaperriere, line 8: "its" state.

src/bsp/include/bsp/ui.h
```diff
@@ -0,0 +1,29 @@
...
+ */
+class UI {
+  public:
+    virtual ~UI(){};
+
+    /**
+     * @brief Provides an interface to print to the console or serial port.
+     * The arguemnts and return values matches the standard printf library
```
stlaperriere, line 16: "arguments"
"match"

src/bsp/include/bsp/ui.h
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * @param [in] format Text to be written, can contain format specifiers that will be replace by
```
stlaperriere, line 18: "replaced"

src/bsp/include/bsp/ui.h
```diff
@@ -0,0 +1,29 @@
...
+     *values specified in the additionnal arguments, matches the standard printf function
+     *
+     * @param [in] ... Additionnal arguments for the format parameter
+     *
+     * @return Matches the standard printf return. The total number of character returned or a
```
stlaperriere, line 23: "characters"

src/logger/include/logger/logger.h
```diff
@@ -0,0 +1,37 @@
...
+ */
+enum class LogLevel { Debug = 0, Info = 1, Warn = 2, Error = 3 };
+
+/**
+ * @brief The return value of the logger
+ */
+enum class LogRet { Ok = 0, LowLevel = 1, Error = 2 };
+
+/**
+ * @brief A logger class with basic logging cappibilities
```
stlaperriere, line 15: "capabilities"

src/logger/include/logger/logger.h
```diff
@@ -0,0 +1,37 @@
...
+class Logger {
+  public:
+    virtual ~Logger(){};
+
+    /**
+     * @brief Logs if the provided level is higher than the current log level
+     *
+     * @param [in] level the log level of the current call
+     *
+     * @param [in] format Text to be written, can contain format specifiers that will be replace by
```
stlaperriere, line 26: "replaced"

src/logger/src/logger_impl.cpp
xgroleau: > In this case, the UI class should be a singleton. We should manage it using an std:shared_pointer. This way, we don't have to do a null check on the parameter passed since it would be caught on the creation of the shared_pointer. Having a shared pointer would also add protection for multithread use. It has relatively low overhead for embedded systems (https://stackoverflow.com/questions/22295665/how-much-is-the-overhead-of-smart-pointers-compared-to-normal-pointers-in-c).

I'm pretty sure share pointers can be null, this doesn't prevent null checking. Whats more is that we will mostly use the stack, not sure what would be the advantage of shared pointer, it would actually be harmfull. Not sure what you mean with multi threaded safety of shared pointer, I think their counter is atomic but thats all. 

What would happen when the dependent goes out of scope? The destructor of the dependency is called on a stack object when it's not even out of it's scope? This would probably be undefined behavior. I agree with the singleton, we could eventually have a static class  or a namespace that manages those dependencies. I'm just not sure how you see this and how it helps here.

g-lesssard:
I currently have build issues on the STM part. My compiler doesn't seem to find cstdarg.h ins ui_impl.cpp. Charles seems to have the same issue. This error is thrown by Clang. My guess is I need a dependency installed, just nor sure what it would be.

This could also be a check to add in the CI: make sure the projects builds on both posix and target. Tests would only be run for posix but at least we have that safety that master will always build and the setup required for building is documented.

xgroleau:
> I currently have build issues on the STM part. My compiler doesn't seem to find cstdarg.h ins ui_impl.cpp. Charles seems to have the same issue. This error is thrown by Clang. My guess is I need a dependency installed, just nor sure what it would be.
> 
> This could also be a check to add in the CI: make sure the projects builds on both posix and target. Tests would only be run for posix but at least we have that safety that master will always build and the setup required for building is documented.

We already have a build for the embedded

src/bsp/include/bsp/ui.h
```diff
@@ -0,0 +1,29 @@
+#ifndef __UI_H_
+#define __UI_H_
+
+/**
+ * @brief Manages the user interface
+ * The user interface can consist of buttons, LED, serial print ports, etc.
+ * It is not a graphical interface.
+ * The board will communicate it's state via the UI and the user can send commands via the UI.
+ */
```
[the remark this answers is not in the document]
xgroleau: > Naming convention: If it is an interface, it should be named it accordingly, iUI or IUI (seems awkward in this case since is short, could be renamed UserInterface for clarity). For the filename, I would also use CamelCase to match what we use for class names.

For the filename, if we go camel case, all the files will go camel case. I have no opinion or preference on that for once, just ask Charles and Sam IRL what they think about it and I will change it if you conclude that camelcase is preferable for the filenames. Just small point though, if we do change, do we change the generated C files too? Or accept that the C files are in snake case and C++ in camelcase?

src/bsp/include/bsp/ui_impl.h
xgroleau: Aight I'll change it. I choosed the Impl way since Impl is only written on definition and once on initialization, then we pass it as the interface. The I prefix need to be everywhere, so you see it "more often". 

Anyway, do we need small confluence page for some conventions like that or it won't even be used? (file naming, interface naming, abstract class naming, etc).

src/logger/include/logger/logger.h
xgroleau: I would avoid mutation if possible, even if it's just a logger. YAGNI,  we'll add the function if we see a use case, but for now let's keep it simple.

src/logger/src/logger_impl.cpp
g-lesssard: My bad on this, I thought that in order to create shared pointers, you had to use std::make_shared, which takes an object and not a pointer in its constructor, making sure we do not have a null reference. You can have empty shared_pointers. :cry: . To respond to the initial comment about null checks in constructors, I would prefer using factories instead of null checks where applicable. 

For thread safety, you are right, only counters are atomic... Read some documentation too fast. We can talk later about how we manage member pointer and some other design standards. If one thing this review has shown is that we need to agree on a baseline of coding standards.

src/bsp/include/bsp/ui.h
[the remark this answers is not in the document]
g-lesssard: True. Either PascalCase or CamelCase, I don't really care, I just dislike snake_case for file names. Again, we need coding standard definition :laughing:

src/bsp/include/bsp/ui_impl.h
g-lesssard: I think a confluence page would be helpful. Settle these conversations outside of a review to concentrate on actual code and less on its looks :smile:

src/freertos/CMakeLists.txt
```diff
@@ -33,8 +32,17 @@ target_include_directories(${LIB_NAME}
 
 target_link_libraries(${LIB_NAME}
     PUBLIC
+    FreeRTOS::Coroutine
+    FreeRTOS::EventGroups
+    FreeRTOS::StreamBuffer
         FreeRTOS::Timers
-        FreeRTOS::Heap::1
+        FreeRTOS::Heap::4
```
g-lesssard, line 39: Quick question: in FreeRTOS::Heap::X, what does the number stands for?

xgroleau: The heap implementation, heap 1 is "static" so freertos has a large buffer and manages it, allows heap on devices that doesn't have one. Heap2 is like heap1 but with some logic to avoir framentation. heap3 is simply a wrapper over malloc and free, so it uses the device heap. Heap4 is like 2 but better at the cost of some memory overhead. Heap5 is like heap4 but I think you can have multiple non-contiguous heap spaces.   We may change to heap3 eventually, or reduce our platform heap to 0. I'm not sure which we should use yet. You can have more info here: https://www.freertos.org/a00111.html

xgroleau: I think we will go for heap3, heap4 has the issue that if we use it on posix, the size_t and the architecture is bigger, thus takes more heap for allocation. In other words, a small heap that works on stm32 doesn't mean it works for posix. We may eventually just encapsulate it in the freertos platform if it becomes an issue

stlaperriere:
Just out of curiosity, I guess that there are probably many existing libs for circular buffers that we could have included. Is there a particular reason for which you decided to implement this yourself?

src/bsp/src/common/include/common/circular_buff.h
```diff
@@ -0,0 +1,178 @@
...
+    const uint8_t* data;
+    /** The length of the available data */
+    const uint16_t length;
+    /** The status of the operation */
+    const CircularBuffRet status;
+} ZeroCopyBuff;
+
+/**
+ *
+ * @brief Initialise the circular buffer struc
```
stlaperriere, line 57: "struct"

xgroleau:
> Just out of curiosity, I guess that there are probably many existing libs for circular buffers that we could have included. Is there a particular reason for which you decided to implement this yourself?

It's a good question actually. For embedded C, I try to avoid using external libs when possible so it plays nice with our build system, have more control on the code size and features, but I definitely inspire myself of what is already done. For example, [I used this guide with a couple changes](https://embeddedartistry.com/blog/2017/05/17/creating-a-circular-buffer-in-c-and-c/). It does give us some more flexibility too, the zerocopy method is an example that does not seem common in libraries and it gives us the freedom to implement it.

src/bsp/include/bsp/IBSP.h
```diff
@@ -0,0 +1,14 @@
+#ifndef __IBSP_H_
+#define __IBSP_H_
+
+class IBSP {
+  public:
+    virtual ~IBSP() = default;
+
+    /**
+     * @brief Initialise the
```
cquesnel, line 9: Initialise the...

src/bsp/src/stm32f4/hal/stm32f429zi/Core/Src/stm32f4xx_it.c
```diff
@@ -0,0 +1,198 @@
...
+    /* USER CODE END USART3_IRQn 0 */
+    HAL_UART_IRQHandler(&huart3);
+    /* USER CODE BEGIN USART3_IRQn 1 */
+
+    /* USER CODE END USART3_IRQn 1 */
+}
+
+/* USER CODE BEGIN 1 */
+
+void HAL_UART_TxCpltCallback(UART_HandleTypeDef* huart) {
```
cquesnel, line 183: I'm not convinced placing this in the stm32f4_it.c file is the way to go. I wouldn't mind putting it in a file with other ISR's but this code is not just valid for the stm32f429zi, it would work for any stm32 and as such, I think it should be a little higher up in the HAL layer.

Same goes for the variables used eg `cbuffUart3`.

src/bsp/src/stm32f4/hal/stm32f429zi/Core/Src/stm32f4xx_it.c
```diff
@@ -0,0 +1,198 @@
...
+    if (huart == &huart3) {
```
cquesnel, line 184: We should use a define instead of `huart3` that way we are less platform dependant.
Same goes for `cbuffUart3`. It could be renamed based on its use: `cbuffDebugUart`

src/bsp/src/stm32f4/hal/stm32f429zi/Core/Src/stm32f4xx_it.c
cquesnel: I think we should decouple ourselves as much as we can from the generated files. It will just make everything easier if we ever have to migrate.

cquesnel (changes requested):

src/bsp/src/common/src/circular_buff.c
```diff
@@ -0,0 +1,164 @@
+#include "common/circular_buff.h"
+#include <stdbool.h>
+#include <stddef.h>
+
+// TODO: Add assert instead of NULL checks
+bool CircularBuff_init(CircularBuff* circularBuff, uint8_t* buffer, uint16_t size) {
+    if (circularBuff == NULL || buffer == NULL) {
```
g-lesssard, line 7: The circular buffer does not manage its own memory? What if the memory to which the buffer refers is deallocated or goes out of scope? It would pass the initial null check but afterwards, run into issues. It may not be problematic the way it is used now, but would be subject to defects if changes occurred. I would either use a malloc (I know, dynamic = bad, but not necessarily illegal) or have a static buffer of a defined length. A check would be made on the size to make sure the size required is not larger than this buffer.  This would avoid having to constantly run null checks of every function.

Also, why C instead of C++ for the circular buff? I don't mind either, just wanted to know if there was a specific reason why.

g-lesssard: Hmm, I guess it doesn't manage the memory locally to have multiple instances of the buffer... If it has to be in C, then I would leave it as is.

g-lesssard (changes requested):

src/bsp/src/common/src/circular_buff.c
xgroleau: How can it manage it's own memory? If you do a malloc in the init, you need to have to call a deinit, which comes back to the same things as not managing memory. This allows the user to manage it's own memory and thus be more flexible. The user can easily write a small wrapper for the init to match you description if needed.

It needs to be in C since it's used in the low level generated C files and may not be only in C++ files.

cquesnel (approved):

src/bsp/src/common/src/circular_buff.c
g-lesssard: True! And makes sense to have it in C then.

g-lesssard (approved):

PR #6, opened 2020-12-29
Feat/ros build

cquesnel:
What was done:

- Make native target compile as a ROS node
- Add example custome ROS message with publish on a topic inside of a FreeRTOS task
- Add Factory for BSP
- Fix bug with STM printf
- Auto-publish documentation to GitHub pages on push to master


src/bsp/src/posix/src/BSP.cpp
```diff
@@ -1,6 +1,64 @@
...
+
+    BSP* bsp = (BSP*)(BSPFactory::getBSP());
+    ros::Publisher publisher =
+        bsp->getRosNodeHandle()->advertise<hive_mind::ExampleMessage>("exampleTopic", 1000);
+
+    while (true) {
+        publisher.publish(msg);
+        msg.number++;
+
+        vTaskDelay(loopRate);
```
stlaperriere, line 47: I see here that you don't use `ros::spin()`. I guess this has to do with the fact that we use FreeRTOS in the HiveMind to orchestrate the tasks. Could this be an issue when we integrate this with the other ROS nodes?

cquesnel: My bad, it got removed at some point during my tests. I've added a call to `ros::spinOnce()` in the `ros_watcher` task, so it is called periodically. It doesn't seem to be causing any issues so far.

The only thing I found that doesn't work is using a sleep directly from ros as in the tutorial, but we can use a `vTaskDelay` without any problems
```
ros::Rate r(10); // 10 hz
while (ros::ok())
{
  ... do some work ...
  r.sleep();
}
```

stlaperriere (changes requested):
There is a problem with clang-tidy when we integrate this build with the SwarmUS-ROS repo. The execution of clang-tidy results in some segmentation faults while parsing the ROS packages. Commenting out the line which includes clang-tidy in the CMakeLists helped me bypass the problem and build the whole project without any other issues.

I spoke with @cquesnel and we found that there seems to be a recurring problem when clang-tidy tries to parse some code that relies on Boost. We have yet to discover the root cause.

src/bsp/src/posix/src/BSP.cpp
```diff
@@ -1,6 +1,65 @@
...
+
+    while (ros::ok()) {
+        ros::spinOnce();
+        vTaskDelay(loopRate);
+    }
+
+    vTaskEndScheduler();
+}
+
+void exampleTopicPublish(void* param) {
```
xgroleau, line 33: I know this is just an example, but maybe make a namespace for our topics to regroup them

src/bsp/src/posix/src/BSP.cpp
```diff
@@ -1,6 +1,65 @@
...
+    (void)param;
+    const int loopRate = 4000;
+    hive_mind::ExampleMessage msg = hive_mind::ExampleMessage();
+    msg.number = 0;
+    msg.text = "Hello World";
+
+    BSP* bsp = (BSP*)(BSPFactory::getBSP());
```
xgroleau, line 40: Pass it via the param, removes the factory dependency. here

stlaperriere (approved):

src/bsp/src/posix/src/UserInterface.cpp
```diff
@@ -1,14 +1,21 @@
...
 
 int UserInterface::print(const char* format, ...) const {
+    int retValue = 0;
 
-    va_list args;
-    va_start(args, format);
-    int retValue = printf(format, args);
-    printf("\n");
-    va_end(args);
+    // ROS logging methods do not return the standard printf value. So let's just return -1 if there
```
xgroleau, line 9: Return the number of character printed, use snprintf, it returns the same thing as printf. You can reuse the string  and pass it to ROS_INFO, preventing to format twice.

src/bsp/src/stm32/hal/src/uart_print.c
```diff
@@ -53,8 +53,9 @@ void UartPrint_sendTxCallback() {
     ZeroCopyBuff buff = CircularBuff_getZeroCopy(&cbuffUartPrint, UINT16_MAX);
     if (buff.status == CircularBuff_Ret_Ok) {
         HAL_UART_Transmit_IT(HUART_PRINT, (uint8_t*)buff.data, buff.length);
-        lastUartPrintTransferSize = buff.length;
     }
+
+    lastUartPrintTransferSize = buff.length;
```
xgroleau, line 58: Is it to prevent retries?

src/freertos/platforms/posix/CMakeLists.txt
```diff
@@ -7,7 +7,11 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 find_package(FreeRTOS COMPONENTS POSIX REQUIRED)
 
-add_library(${LIB_NAME} INTERFACE)
+set(LIB_SOURCES
+    platform_freertos_hooks.c
+)
+
+add_library(${LIB_NAME} STATIC ${LIB_SOURCES})
```
xgroleau, line 14: I would keep it as an interface just to make sure all symbols are resolved properly, target_sources should do the trick

xgroleau (changes requested):
GJ, it's all coming together pretty well and ROS is nicely encapsulated. Just a couple suggestions and nitpicks.

cmake/catkin/common.cmake
```diff
@@ -0,0 +1,32 @@
...
+        message_generation
+        )
+
+    add_message_files(
+        DIRECTORY ros/msg
+        FILES
+        ${ROS_MESSAGES_LIST}
+    )
+
+   #TODO: Uncomment if service_generation is needed
```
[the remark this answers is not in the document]
cquesnel: Not yet, but it will be at some point in the near future, I would leave it there until then

src/bsp/src/posix/src/BSP.cpp
```diff
@@ -1,6 +1,65 @@
...
+
+        vTaskDelay(loopRate);
+    }
+}
+
+void BSP::initChip(void* args) {
+    CmdLineArgs* cmdLineArgs = (CmdLineArgs*)args;
+    ros::init(cmdLineArgs->m_argc, cmdLineArgs->m_argv, "hive_mind");
+
+    m_rosNodeHandle = new ros::NodeHandle();
```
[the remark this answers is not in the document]
cquesnel: I wanted to, but it's impossible. The nodeHandle must be created after the call to `ros::init()`

src/bsp/src/posix/src/BSP.cpp
```diff
@@ -1,6 +1,65 @@
...
+void rosWatcher(void* param) {
+    (void)param;
+    const int loopRate = 100;
+
+    while (ros::ok()) {
+        ros::spinOnce();
+        vTaskDelay(loopRate);
+    }
+
+    vTaskEndScheduler();
```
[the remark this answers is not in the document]
cquesnel: ros is not ok when the node is killed. At this point, ros isn't running anymore and the log would therefore not do anything.
Not sure what you mean by "pass a necessary object via the param"

src/bsp/src/stm32/hal/src/uart_print.c
cquesnel: It's to fix a bug where if there is a problem while getting the buffer, the lastUartPrintSize never returns to 0 and the circular buffer only returns errors from thereon

src/bsp/include/bsp/BSPFactory.h
```diff
@@ -0,0 +1,18 @@
+#ifndef _BSPFACTORY_H
+#define _BSPFACTORY_H
+
+#include "bsp/IBSP.h"
+#include "bsp/IUserInterface.h"
+
+class BSPFactory {
```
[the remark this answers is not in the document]
cquesnel: Changed for namespace and renamed to `BSPContainer`.

src/Main.cpp
```diff
@@ -24,15 +25,15 @@ void printThreadExample(void* param) {
     }
 }
 
-int main() {
-    BSP bsp = BSP();
-    bsp.initChip();
+int main(int argc, char** argv) {
+    CmdLineArgs cmdLineArgs = {argc, argv};
```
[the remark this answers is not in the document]
cquesnel: argc: 536874216 (I would guess some random value in RAM)
argv[0][0]: '\0'

src/bsp/src/posix/src/BSP.cpp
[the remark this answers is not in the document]
xgroleau: ahhh I see, that explains the pointer.

src/bsp/src/posix/include/BSP.h
```diff
@@ -0,0 +1,19 @@
...
+class BSP : public IBSP {
+  public:
+    BSP();
+    ~BSP() override;
+
+    void initChip(void* args) override;
+    ros::NodeHandle* getRosNodeHandle();
+
+  private:
+    ros::NodeHandle* m_rosNodeHandle;
```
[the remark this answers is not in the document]
cquesnel: Changed for shared_ptr

src/bsp/src/posix/src/BSP.cpp
cquesnel: Made all topics relative to the node name (see `new ros::NodeHandle` in `BSP::initChip()`)

cmake/googletest/common.cmake
```diff
@@ -7,10 +7,14 @@ function(googletest_get_populate)
     GIT_TAG        release-1.10.0
     )
 
-    FetchContent_GetProperties(googletest)
-    if(NOT googletest_POPULATED)
-        FetchContent_Populate(googletest)
-        add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR})
+    if (TARGET gtest)
```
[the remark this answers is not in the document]
cquesnel: I changed things around so get_populate only fetches GTest, but I had to continue using `if (TARGET gtest)` instead of `find_package(GTest)` as the second option makes the LGTM build fail.

src/bsp/include/bsp/BSPContainer.h
```diff
@@ -0,0 +1,14 @@
...
+#define _BSPFACTORY_H
+
+#include "bsp/IBSP.h"
+#include "bsp/IUserInterface.h"
+
+namespace BSPContainer {
+/**
+ * @brief Return an instance of the platform dependent BSP.
+ */
+IBSP* getBSP();
```
xgroleau, line 11: Just realized something, change for a reference would be more appropriate I think, I know it's rare to return a reference, but it makes it explicit that the Container manages the memory and I think it's legitimate in this singleton case.

cquesnel: Good idea

cmake/googletest/common.cmake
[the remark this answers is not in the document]
xgroleau: Fair enough for me

xgroleau (approved):
LGTM :+1:

package.xml
```diff
@@ -0,0 +1,30 @@
...
+  <build_depend>std_msgs</build_depend>
+  <build_depend>message_generation</build_depend>
+
+  <build_export_depend>roscpp</build_export_depend>
+  <build_export_depend>rospy</build_export_depend>
+  <build_export_depend>std_msgs</build_export_depend>
+  <build_export_depend>message_runtime</build_export_depend>
+
+  <exec_depend>roscpp</exec_depend>
+  <exec_depend>rospy</exec_depend>
```
g-lesssard, line 26: Remove dependencies to rospy

g-lesssard (commented):
Good job

g-lesssard (approved):

PR #7, opened 2021-01-02
Feat/bittybuzz support

xgroleau:
- Added support for bittybuzz
- Added support for compiling bzz file in .h with bytecode
- Fixed logger not releasing the mutex


src/bittybuzz/include/bittybuzz/BittyBuzzVm.h
```diff
@@ -0,0 +1,42 @@
...
+    bbzvm_funp functionPtr;
+
+} FunctionRegister;
+
+class BittyBuzzVm : public IBittyBuzzVm {
+  public:
+    BittyBuzzVm(const IBittyBuzzBytecode& bytecode,
+                const IBSP& bsp,
+                const ILogger& logger,
+                const FunctionRegister* functionRegisters,
```
xgroleau, line 21: Tried to find a way to pass a iterator, similar to IEnumerator in C# but it would require template, so code in the .h. suggestions are appreciated

cquesnel: I mean templating is an integral part of C++ and using it requires some code in the .h so at that point I don't mind using it. It's not like if we had all of the code in the .h

src/bittybuzz/src/BittyBuzzVm.cpp
```diff
@@ -0,0 +1,46 @@
...
+    m_bytecode(bytecode), m_bsp(bsp), m_logger(logger) {
+    // Init global variable
+    vm = &m_bbzVm;
+    BittyBuzzSystem::logger = &logger;
+
+    // Init vm
+    bbzvm_construct(m_bsp.getUUId());
+    bbzvm_set_error_receiver(BittyBuzzSystem::errorReceiver);
+    bbzvm_set_bcode(m_bytecode.getBytecodeFetchFunction(), m_bytecode.getBytecodeLength());
+    bbzringbuf_construct(&m_bbzPayloadBuff, m_bbzMsgBuff, 1, 11);
```
cquesnel, line 22: This is the second time I see the number 11 somewhere. What is it? And could we maybe move it to a define?

cquesnel (changes requested):
GJ! I'm sure @g-lesssard will have more to say about the BBVM integration as he has already worked with the BVM, but on my part only a few comments.

src/Main.cpp
```diff
@@ -13,20 +16,26 @@ void printThreadExample(void* param) {
     (void)param;
     const int toggleDelay = 2;
 
+    BSP bsp = BSP();
+    bsp.initChip();
```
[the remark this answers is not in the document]
xgroleau: It's  temporary until I have your changes with the BSP container so I could test

src/bittybuzz/src/BittyBuzzVm.cpp
xgroleau: Raised it to 16, added a macro and changed to an std::array instead

src/Main.cpp
```diff
@@ -13,20 +16,26 @@ void printThreadExample(void* param) {
...
+
     UserInterface ui = UserInterface();
     Logger logger = Logger(LogLevel::Debug, ui);
 
+    BittyBuzzBytecode bytecode = BittyBuzzFactory::createBittyBuzzBytecode(logger);
+    auto functionRegisters = BittyBuzzFactory::createBittyBuzzFunctionRegisters();
```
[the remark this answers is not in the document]
xgroleau: I though we said we would avoid auto. Thing is that the type here has the length of the std::array, just wanted to avoid changing it at two places at the same time. I'll change it.

src/bittybuzz/include/bittybuzz/BittyBuzzVm.h
xgroleau: I've added a .tpp file so it's still isolated in another file for readability

stlaperriere (commented):
:eyes:

cquesnel (approved):

g-lesssard (approved):
Good job!

src/bittybuzz/include/bittybuzz/BittyBuzzVm.tpp
```diff
@@ -0,0 +1,34 @@
+
```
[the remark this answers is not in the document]
xgroleau: We needed a template and didn't want to have a lot of code in .h, so I created a .tpp file. So it has the implementation of the template

PR #9, opened 2021-01-07
Remove documentation from not compiled targets

cquesnel:


xgroleau (approved):
We could have used doxygen_add_doc instead of having an if and exluding the non-current one. But it's fine since it's not like we have 5 different BSP

PR #10, opened 2021-01-09
Feat/bittybuzz string support

xgroleau:
Added string support for the bbvm. On compilation a header file is created via a bash script. The symbols in the bst file are ignored since we know those are function names. 

I've added a verification to check if the ID stored and requested match. Hence why it's an std::pair<uint16_t, char*> and not just an array of char*. Not sure if it's really needed, but it makes it easier to find if the build breaks since it's only a bash script. It does add 2 bytes of overhead (uint16_t) for each string.


g-lesssard (approved):
LGTM!

src/bittybuzz/tests/CMakeLists.txt
```diff
@@ -18,26 +20,38 @@ set(TEST_HEADERS
 
 # bzz target definitions
 set(BZZ_TEST_SCRIPTS_LOCATION ${CMAKE_CURRENT_SOURCE_DIR}/vm_tests/buzz_tests_scripts)
+
+set(LOGINT_BST ${BITTYBUZZ_HIVEMIND_BST} vm_tests/buzz_tests_scripts/logInt.bst)
+list(LENGTH LOGINT_BST TEST)
+message("WOOW ${LOGINT_BST}")
```
cquesnel, line 26: Is this still necessary?

cquesnel (approved):
Approved. Only two debug prints to remove in a CMake

stlaperriere (approved):
This will be very useful I think. GJ!

src/bittybuzz/tests/CMakeLists.txt
xgroleau: Good catch ahaha

PR #12, opened 2021-01-14
Added pheromones to build

xgroleau:
Added pheromones to buildsystem


cmake/pheromones/common.cmake
```diff
@@ -0,0 +1,43 @@
...
+
+function(pheromones_fetch_populate)
+    if(PHEROMONES_PATH)
+        message(INFO "PHEROMONES_PATH specified, skipping fetch")
+    endif()
+
+    FetchContent_Declare(
+        pheromones
+        GIT_REPOSITORY https://github.com/SwarmUS/Pheromones
+        GIT_TAG        01bd792711588c425fe0d5e540cbdcff746f5fec
```
g-lesssard, line 15: Maybe change to branch master instead of tag? To make sure to always have the newest version on the build.

cmake/pheromones/common.cmake
```diff
@@ -0,0 +1,43 @@
...
+        GIT_PROGRESS   TRUE
+    )
+    set(PHEROMONES pheromones)
+    string(TOLOWER ${PHEROMONES} PHEROMONES_L)
+
+    FetchContent_GetProperties(PHEROMONES POPULATED PHEROMONES_POPULATED)
```
g-lesssard, line 21: Change PHEROMONES to ${PHEROMONES} to properly resolve the POPULATED property. The same bug occurs is present in the freertos common.cmake. can you fix it at the same time?

g-lesssard (changes requested):

g-lesssard (changes requested):
Small fixes needed

stlaperriere (approved):

cmake/pheromones/common.cmake
xgroleau: I would maybe let it a tag, if not, updating pheromones could break your build. Ex during a Hivemind PR your build could break because pheromones was updated

g-lesssard (approved):

PR #13, opened 2021-01-18
Feat/swarinfo 147 driver socket ros

xgroleau:
Added socket build for POSIX build, stm32 will come eventually


src/bsp/include/bsp/SocketContainer.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include "ITCPClient.h"
+#include "TCPClientWrapper.h"
+#include <logger/ILogger.h>
+#include <optional>
+
+/**@brief Containes the sockets for the application*/
+namespace SocketContainer {
+    /**
+     * @brief Return an instance of the platform dependent BSP.
```
cquesnel, line 12: Wrong comment

src/bsp/include/bsp/SocketContainer.h
```diff
@@ -0,0 +1,22 @@
...
+     *
+     * @b Warning note that if you make multiple request, you will obtaine multiple TCPClientWrapper
+     * objects, but under the hood, it's the same socket reference
+     */
+    std::optional<TCPClientWrapper> getHostClientSocket(const char* address,
+                                                        int port,
```
cquesnel, line 18: I would maybe change specify the size of the `int` (we are compiling on 64 bit linux and 32 bit STM32)

src/bsp/include/bsp/TCPClientWrapper.h
```diff
@@ -0,0 +1,30 @@
+#ifndef __TCPClientWrapperWRAPPER_H_
+#define __TCPClientWrapperWRAPPER_H_
+
+#include "bsp/ITCPClient.h"
+#include <logger/ILogger.h>
+
+/**
+ *@brief A TCP client wrapper that is as a common socket between platforms.
+ **/
+class TCPClientWrapper : public ITCPClient {
```
cquesnel, line 10: What is the goal of the `TCPClientWrapper` considering it is only returns calls from the `TCPClient` directly? I know it's probably some obvious OOP way of doing things, but it's been a long time since I've used OOP...

src/bsp/src/posix/src/SocketContainer.cpp
```diff
@@ -0,0 +1,22 @@
...
+                                                                     const ILogger& logger) {
+
+    static std::optional<TCPClient> s_clientSocket = {};
+    static std::optional<TCPClientWrapper> s_wrapper = {};
+
+    if (!s_clientSocket) {
+        std::optional<TCPClient> socket = SocketFactory::createTCPClient(address, port, logger);
+
+        if (socket) {
+            s_clientSocket.emplace(socket.value());
```
cquesnel, line 16: I suppose this is the awkward stuff for the `optional` you were talking about?

src/bsp/include/bsp/SocketContainer.h
```diff
@@ -0,0 +1,22 @@
...
+#include "TCPClientWrapper.h"
+#include <logger/ILogger.h>
+#include <optional>
+
+/**@brief Containes the sockets for the application*/
+namespace SocketContainer {
+    /**
+     * @brief Return an instance of the platform dependent BSP.
+     *
+     * @b Warning note that if you make multiple request, you will obtaine multiple TCPClientWrapper
```
stlaperriere, line 14: 'obtain

cquesnel (changes requested):

src/bsp/include/bsp/TCPClientWrapper.h
xgroleau: The  TCPClient is a platform dependent implementation, honestly that's the only reason, and I cannot return a ITCPClient here because optional is a template and does not support reference or incomplete types.

src/bsp/src/posix/src/SocketContainer.cpp
xgroleau: Yep, that and the TCPWrapper stuff, since I can't return a std::optional of an incomplete type or a reference.
though a reference could be fixed with std::referece_wrapper, but still doesn't solve the incomplete type and it starts to get tedious typing

cquesnel (approved):

stlaperriere (approved):

src/Main.cpp
```diff
@@ -19,6 +20,11 @@ void printThreadExample(void* param) {
     UserInterface ui = UserInterface();
     Logger logger = Logger(LogLevel::Debug, ui);
 
+    auto socket = SocketContainer::getHostClientSocket("127.0.0.1", 5555, logger);
```
g-lesssard, line 23: Similarly to the PR for the robot counterpart, the port for the socket should be supplied by the launch file. I would add it now to simplify the integration in the future. Also, the container functions should not take any parameters in my opinion. The container function should fetch all dependencies and configuration necessary.

g-lesssard (changes requested):

src/bsp/include/bsp/SocketContainer.h
```diff
@@ -0,0 +1,20 @@
...
+#include "TCPClientWrapper.h"
+#include <logger/ILogger.h>
+#include <optional>
+
+/**@brief Containes the sockets for the application*/
+namespace SocketContainer {
+    /**
+     * @brief Return an instance of the TPCClient wrapper that contains a platform dependent socket.
+     *
+     */
```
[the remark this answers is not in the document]
xgroleau: Yes we excpect failure on socket initialization, can't know at boot time if the ethernet cable is connected (or the uart cable). This allows will allow us to loop between the both of them and see which one can be initialize, thus choosing our comm interface with the host

cmake/freertos/common.cmake
```diff
@@ -7,9 +7,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} PARENT_SCOPE)
 function(freertos_fetch_kernel)
     FetchContent_Declare(
         FreeRTOSKernel
-        GIT_REPOSITORY https://github.com/FreeRTOS/FreeRTOS-Kernel
-        GIT_TAG        V10.4.1-kernel-only
-        GIT_PROGRESS   TRUE
+        URL https://github.com/FreeRTOS/FreeRTOS-Kernel/archive/V10.4.1-kernel-only.tar.gz
     )
```
g-lesssard, line 11: I like this. A good permanent solution to avoid this problem occurring once again in the future.

src/bsp/src/posix/src/TCPClient.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <cstdio>
+#include <netdb.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+TCPClient::TCPClient(int socket, sockaddr_in address, const ILogger& logger) :
+    m_logger(logger), m_socketFd(socket), m_address(address) {}
+
```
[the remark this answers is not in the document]
xgroleau: Good call, was  for experimentation, fixed it. That's dangerous to let it there

src/Main.cpp
xgroleau: Fixed, inject using ros node handle, idk how to add those args though, we don't have a launch file here, it's in Swarmus-ROS right?

g-lesssard (approved):

PR #14, opened 2021-01-18
Fix/swarinfo 151 fix build system for melodic

xgroleau:
Fix catkin build on ubuntu 18.04 with melodic


g-lesssard (approved):
Nice that the fix is not too complicated. In the future, we could add a melodic build on github in this repo to validate the build, and in SwarmUS-ROS that distributes it.

etienn8 (approved):
I confirm that it works on my PC with Ubuntu 18.04 with Melodic as the ROS distro.

PR #15, opened 2021-01-21
Feat/swarinfo 128 stm uart driver

cquesnel:
Implemented the UART driver for STM.

- Added a `HostUart` class and a `CRC` class as we are using a hardware CRC on the STM
- Compiles for POSIX, but only the strict minimal for compiling is implemented
- Prints when a message is received. Later on this will be changed for a stream based `receive()` function or a subscriber model depending on what we decide to do.


src/bsp/src/stm32/include/HostUart.h
```diff
@@ -0,0 +1,49 @@
...
+    ICRC& m_crc;
+    ILogger& m_logger;
+
+    TxState m_txState;
+    uint8_t m_txLength;
+    uint8_t m_txHeader[HOST_UART_HEADER_LENGTH];
+    const uint8_t* m_txBuffer;
+
+    uint8_t m_rxBuffer[HOST_UART_MAX_MESSAGE_LENGTH];
+    uint8_t m_rxHeader[HOST_UART_HEADER_LENGTH];
```
xgroleau, line 37: std::array, easier to get the length after instead of using the defines (and avoiding the mistake of using the wrong define)

src/bsp/src/stm32/include/HostUart.h
```diff
@@ -0,0 +1,49 @@
...
+    bool isBusy() override;
+
+    void process();
+
+    friend void hostUart_C_txCpltCallback(void* hostUartInstance);
+    friend void hostUart_C_rxCpltCallback(void* hostUartInstance);
+
+  private:
+    enum RxState { Rx_Idle, Rx_WaitForHeader, Rx_WaitForPayload, Rx_CheckIntegrity };
+    enum TxState { Tx_Idle, Tx_SendHeader, Tx_SendPayload };
```
xgroleau, line 26: use `enum class` instead, they don't overlap eachother, or if a global enum has the same name

src/bsp/src/stm32/src/HostUart.cpp
```diff
@@ -0,0 +1,137 @@
...
+        m_logger.log(LogLevel::Warn, "Could not send message to host over UART");
+    }
+
+    return ret;
+}
+
+void HostUart::startHeaderListen() {
+    if (xSemaphoreTake(m_uartSemaphore, (TickType_t)10) == pdTRUE) {
+        UartHost_receiveDMA(m_rxHeader, HOST_UART_HEADER_LENGTH,
+                            (void (*)(void*))(&hostUart_C_rxCpltCallback), (void*)this);
```
xgroleau, line 75: yeaaa a typedef would help this casting and the others under it

src/bsp/src/stm32/src/CRC.cpp
```diff
@@ -0,0 +1,20 @@
+#include "CRC.h"
+#include "hal/hal.h"
+
+CRC::CRC() {
+    m_semaphore = xSemaphoreCreateBinary();
+
+    xSemaphoreGive(m_semaphore);
```
xgroleau, line 7: No null check (though idk what you could do about it), maybe just add a todo to make static

src/bsp/src/stm32/src/HostUart.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    while (true) {
+        static_cast<HostUart*>(param)->process();
+        vTaskDelay(loopRate);
+    }
+}
+
+HostUart::HostUart(ICRC& crc, ILogger& logger) :
+    m_crc(crc), m_logger(logger), m_txState(TxState::Tx_Idle), m_rxState(RxState::Rx_Idle) {
+    m_uartSemaphore = xSemaphoreCreateBinary();
```
xgroleau, line 27: add a todo to make static

xgroleau (changes requested):
Good work! couple comments but most of them a just syntax or stuff like that, no real big issue.

src/bsp/src/stm32/src/HostUart.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <cstring>
+#include <task.h>
+
+void hostUart_C_txCpltCallback(void* hostUartInstance) {
+    static_cast<HostUart*>(hostUartInstance)->txCpltCallback();
+}
+
+void hostUart_C_rxCpltCallback(void* hostUartInstance) {
+    static_cast<HostUart*>(hostUartInstance)->rxCpltCallback();
+}
```
xgroleau, line 14: Maybe static since it's limited to the scope of this file

src/bsp/src/stm32/src/HostUart.cpp
```diff
@@ -0,0 +1,137 @@
...
+    default:
+        break;
+    }
+}
+
+void HostUart::process() {
+    if (m_rxState == RxState::Rx_CheckIntegrity) {
+        uint32_t calculatedCrc = m_crc.calculateCRC32(m_rxBuffer, m_rxLength);
+        if (calculatedCrc == m_rxCrc) {
+            m_logger.log(LogLevel::Info, "Received UART message of %d bytes", m_rxLength);
```
xgroleau, line 126: Probably a debug level, if not we will have a lot of spamming in info

src/bsp/src/stm32/include/HostUart.h
xgroleau: Though I saw you are using :: notation for access, maybe it's already an enum class since it's in defined in a class

src/bsp/include/bsp/IHostUart.h
```diff
@@ -0,0 +1,32 @@
+#ifndef __IHOSTUART_H__
+#define __IHOSTUART_H__
+
+#include <cstdint>
+
+#define HOST_UART_MAX_MESSAGE_LENGTH UINT16_MAX
+#define HOST_UART_HEADER_LENGTH 6
```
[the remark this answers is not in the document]
cquesnel: Removed in HostUart.h so it can be shared between both targets

src/bsp/src/stm32/hal/src/hal.c
```diff
@@ -12,8 +12,19 @@ void Hal_init() {
...
     /* Initialize UartPrint */
     UartPrint_init();
-}
\ No newline at end of file
+}
+
+uint32_t Hal_calculateCRC32(const uint8_t* buffer, uint32_t length) {
+    // TODO: Determine what we do if we have a buffer that is not a multiple of 32 bytes
+    // The HAL_CRC function takes the number of 32 bit words and not the number of bytes
+    return HAL_CRC_Calculate(&hcrc, (uint32_t*)buffer, length >> 2);
```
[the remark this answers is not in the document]
cquesnel: A CRC32 works on blocks of 4 bytes, so if we calculate the CRC on something that isn't a multiple of 4 bytes, the overflow bytes are just not considered in the calculation. I will have to check how the Java implementation behaves and from there, we can decide how we will handle that case (I suggest padding with zeroes to reach a multiple of 4 bytes, but we will see)

src/bsp/src/stm32/hal/src/uart_host.c
```diff
@@ -0,0 +1,54 @@
...
+        txCpltCallbackFct = cpltCallback;
+        txCallbackInstance = instance;
+        if (HAL_UART_Transmit_DMA(HUART_HOST, buffer, length) != HAL_OK) {
+            ret = false;
+        }
+    } else {
+        ret = false;
+    }
+
+    return ret;
```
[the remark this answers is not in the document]
cquesnel: I've used to never return before the end of a function, but if it's fine with you, will change it.

src/bsp/include/bsp/IHostUart.h
```diff
@@ -0,0 +1,32 @@
...
+
+class IHostUart {
```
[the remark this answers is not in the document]
cquesnel: Added for HostUart, CRC and Logger

src/bsp/src/stm32/include/HostUart.h
cquesnel: Changed for enum class

src/bsp/src/stm32/src/HostUart.cpp
cquesnel: Doesn't work if we make it a friend of the HostUart class. Leaving it as is.

src/bsp/src/stm32/src/HostUart.cpp
cquesnel: Was planning on removing it when we have a way to pass messages up to higher level, but will change

xgroleau (approved):

PR #17, opened 2021-01-25
Feat/swarinfo 130 driver uart ros

cquesnel:
Integrated the TCP server to mock the uart connection to the phone.

- Client reconnection is not handled. We may choose to do it later, but it is out of scope for the moment
- Added parameter in launch file. Can set the port to 0 so the server is not opened (Will be useful when we have multiple hiveboards in the same simulation)


src/bsp/src/posix/src/TCPUartMock.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    m_logger.log(LogLevel::Info, "UART TCP mock server waiting for client on port %d", m_port);
+
+    xTaskCreate(TCPUartMock_listenTask, "tcp_uart_mock_listen", configMINIMAL_STACK_SIZE * 100,
+                (void*)this, tskIDLE_PRIORITY + 1, NULL);
+}
+
+bool TCPUartMock::send(const uint8_t* buffer, uint16_t length) {
+    if (!m_hasClient) {
```
xgroleau, line 51: is it possible to use the client file descriptor instead of using 2 states?  i.e check if the file descriptor is != -1 for example

Or even an std::optional<int> m_clienFd

src/bsp/src/posix/src/TCPUartMock.cpp
```diff
@@ -0,0 +1,101 @@
...
+        return false;
+    }
+
+    return ::send(m_clientFd, buffer, length, 0) >= 0;
```
xgroleau, line 55: Verify the length

` ::send(m_clientFd, buffer, length, 0) == (int)length?`

src/bsp/src/stm32/src/HostUart.cpp
```diff
@@ -70,6 +70,8 @@ bool HostUart::send(const uint8_t* buffer, uint16_t length) {
     return ret;
 }
 
+int32_t HostUart::receive(uint8_t* buffer, uint16_t length) const { return 0; }
```
xgroleau, line 73: Add //TODO:

src/bsp/src/posix/src/BSP.cpp
```diff
@@ -50,7 +54,11 @@ void BSP::initChip(void* args) {
 
     m_rosNodeHandle = std::shared_ptr<ros::NodeHandle>(new ros::NodeHandle("~"));
 
-    xTaskCreate(rosWatcher, "ros_watch", configMINIMAL_STACK_SIZE, NULL, tskIDLE_PRIORITY + 1,
+    TCPUartMock& tcpUart = static_cast<TCPUartMock&>(BSPContainer::getHostUart());
+    int port = m_rosNodeHandle->param("uart_mock_port", 0);
+    tcpUart.openSocket(port);
```
xgroleau, line 59: Would this belong in the container? 
You can get the node handle from the bsp in the container

You could use an [std::call_once](https://en.cppreference.com/w/cpp/thread/call_once) just after the static variable declaration

src/bsp/src/posix/src/BSP.cpp
```diff
@@ -23,6 +24,9 @@ void rosWatcher(void* param) {
         vTaskDelay(loopRate);
     }
 
+    TCPUartMock& uartMock = static_cast<TCPUartMock&>(BSPContainer::getHostUart());
+    uartMock.close();
```
xgroleau, line 28: Is it needed since it's called in the destructor? Idk, i'm legit curious for a static object if it's destructor is called

xgroleau (changes requested):
GJ, like the wrapper for disabling signals. Couple comments and questions

src/bsp/src/posix/src/BSP.cpp
```diff
@@ -50,7 +54,11 @@ void BSP::initChip(void* args) {
 
     m_rosNodeHandle = std::shared_ptr<ros::NodeHandle>(new ros::NodeHandle("~"));
 
-    xTaskCreate(rosWatcher, "ros_watch", configMINIMAL_STACK_SIZE, NULL, tskIDLE_PRIORITY + 1,
+    TCPUartMock& tcpUart = static_cast<TCPUartMock&>(BSPContainer::getHostUart());
+    int port = m_rosNodeHandle->param("uart_mock_port", 0);
```
t-constantin, line 58: Is it better to define "uart_mock_port" somewhere else than hard coded? The comment goes for "hive_mind" at line 53 as well

src/bsp/src/posix/src/TCPUartMock.cpp
```diff
@@ -0,0 +1,101 @@
...
+    if (ret == 0) {
+        // TODO: If we ever want to handle client reconnection in the simulation, do it here
+        m_logger.log(LogLevel::Warn,
+                     "Error while reading UART socket. Client has probably disconnected");
+    }
+
+    return ret;
+}
+
+bool TCPUartMock::isBusy() const { return false; }
```
t-constantin, line 73: If always false, still relevant?

t-constantin (approved):
Pretty clean code!
My comments are very minor, I won't be blocking the review for that

src/bsp/src/posix/src/BSP.cpp
cquesnel: This is the only place where I've managed to get it working every time. If I put it in the container, it doesn't always work. Also, I think it makes sense that the socket is opened as part of the init chip routine, it's part of the initialisation process. Imagine if we don't do a get() on the driver, the socket isn't opened and the phone can't connect.

src/bsp/src/posix/src/BSP.cpp
cquesnel: yeah i think it's called in the destructor, I was just testing things as ROS had problems shutting down the node.

src/bsp/src/posix/src/TCPUartMock.cpp
cquesnel: Yes because it's used in the STM version of the driver. So it has to be here for compatibility reasons.

src/bsp/src/posix/src/BSP.cpp
cquesnel: @xgroleau We should have a string factory somewhere. Do we put that in the main cleanup task?

src/bsp/src/posix/src/TCPUartMock.cpp
cquesnel: Changed for std::optional

src/bsp/src/posix/src/BSP.cpp
xgroleau: Or more like getters in the ros BPS implementation? But yea that could go in the main cleanup

xgroleau (approved):

PR #22, opened 2021-02-03
Feat/swarinfo 139 add robot ID

cquesnel:
Created a settings manager that recovers settings from:
- CMake generate file for STM build
- RosParams for ROS build

Code is currently duplicated between both SocketContainers, but fixing that is out of scope for this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,6 @@ openocd -f ./tools/openocd/stm32_f4/stm32_f4.cfg -c init -c \"reset init\"
 ### External connections
 #### Ethernet port
 The firmware assigns a static IP of 192.168.1.10 to the device with a subnet mask of 255.255.255.0.
-The host computer is expected to have the IP 192.168.1.101 (WILL BE CONFIGURABLE LATER ON) and a TCP socket
-opened on port 5555.
 
 The ethernet speed can be measured with [iperf 2](https://iperf.fr/iperf-download.php).
 To open the server on the device, simply build with the `ENABLE_TARGET_IPERF_SERVER` CMake option:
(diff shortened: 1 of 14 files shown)


src/bsp/include/bsp/SettingsContainer.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include <stdint.h>
+#include <string>
+
+namespace SettingsContainer {
+    uint16_t GetUUID();
+
+    uint32_t GetHostPort();
+
+    std::string GetHostIP();
```
xgroleau, line 12: Maybe we should use a const char* since std::string uses heap

src/bsp/src/stm32/src/BSP.cpp
```diff
@@ -10,7 +11,4 @@ void BSP::initChip(void* args) {
     Hal_init();
 }
 
-uint16_t BSP::getUUId() const {
-    // TODO: Change do the ID is obtained from persistent memory
-    return 1;
-}
\ No newline at end of file
+uint16_t BSP::getUUId() const { return SettingsContainer::GetUUID(); }
```
xgroleau, line 14: Just to be sure, the plan here is to eventually use the UUID from the EPROM and if there is none, we will store and use the UUID from the settings container?

README.md
```diff
@@ -119,6 +117,21 @@ The firmware has been configured with the following pins for the uart connection
...
-#### BittyBuzz
-* Extern C on generated .h file
+### Configuration
+Different firmware settings can be configured at build time (for the STM32 target) or runtime (for the ROS target).
+
+#### STM32 Target
+CMake variables can be used to override certain default firmware settings.
+| Variable  | Default value   |
+| ---       | ---             | 
+| UUID      | 1               |
```
xgroleau, line 127: Maybe just mention that the UUID of 0 is reserved for broadcast/invalid

src/bsp/src/stm32/src/SocketContainer.cpp
```diff
@@ -11,13 +12,11 @@ std::optional<TCPClientWrapper> SocketContainer::getHostClientSocket() {
     if (!s_clientSocket) {
 
         ILogger& logger = LoggerContainer::getLogger();
+        std::string hostAddress = SettingsContainer::GetHostIP();
```
xgroleau, line 15: Don't forget to change here for a const char* after the change on GetHostIp so you don't use a std::string constructor

xgroleau (changes requested):
GJ! minor details for std::string

src/bsp/src/stm32/src/BSP.cpp
cquesnel: Exactly

xgroleau (approved):
LGMT 👍
GJ on using snprintf (which is safe) instead of usafe strcpy

PR #23, opened 2021-02-04
Refactor for using stack

xgroleau:


src/bsp/src/stm32/src/SocketFactory.cpp
```diff
@@ -23,6 +23,7 @@ std::optional<TCPClient> SocketFactory::createTCPClient(const char* address,
     // Connect the socket
     if (lwip_connect(sockfd, (sockaddr*)&serverAddr, sizeof(serverAddr)) < 0) {
         logger.log(LogLevel::Info, "Could not connect to server %s : %d", address, port);
+        lwip_close(sockfd);
```
cquesnel, line 26: Do we need to do the same thing on the POSIX version?

src/bsp/src/posix/include/TCPUartMock.h
```diff
@@ -24,6 +25,8 @@ class TCPUartMock : public IHostUart {
   private:
     ILogger& m_logger;
 
+    BaseTask<configMINIMAL_STACK_SIZE * 100> m_listenTask;
```
cquesnel, line 28: Was the x100 a test or do we really need such a big stack?

cquesnel (approved):

cquesnel (approved):
GJ! Very clean with the tasks and especially the LockGuard

src/bsp/src/stm32/src/SocketFactory.cpp
xgroleau: hmmmm good question, I'll check it out

src/bsp/src/posix/include/TCPUartMock.h
xgroleau: It was the value that was there when I refactored it. I'll change it to something smaller

cquesnel: my bad, it was probably a test from me

src/bsp/src/stm32/src/SocketFactory.cpp
xgroleau: Posix socket are now closed on fail

PR #24, opened 2021-02-07
Feat/swarinfo 152 main cleanup

xgroleau:
Small cleanup of the main


src/bsp/src/stm32/src/SettingsContainer.cpp
```diff
@@ -2,10 +2,12 @@
...
+uint32_t SettingsContainer::getHostPort() { return HOST_PORT; }
 
-uint8_t SettingsContainer::GetHostIP(char* buf, uint8_t length) {
+uint8_t SettingsContainer::getHostIP(char* buf, uint8_t length) {
     return snprintf(buf, length, HOST_IP);
-}
\ No newline at end of file
+}
+
+LogLevel SettingsContainer::getLogLevel() { return LOG_LEVEL; }
```
cquesnel, line 13: In the ROS version we allow `Err` and `Warning` while here we don't. Maybe remove the extra supported strings from the ROS version

src/bsp/src/posix/src/SettingsContainer.cpp
```diff
@@ -9,23 +9,43 @@ std::shared_ptr<ros::NodeHandle> getRosNodeHandle() {
...
+        return LogLevel::Info;
+    }
+    if (logLevel == "Warn" || logLevel == "Warning") {
+        return LogLevel::Warn;
+    }
+    if (logLevel == "Error" || logLevel == "Err") {
+        return LogLevel::Error;
+    }
+
+    return LogLevel::Debug;
```
cquesnel, line 50: Considering our default is `Info`. We should maybe return info here

src/bsp/src/stm32/include/DefaultSettings.in.h
```diff
@@ -1,6 +1,9 @@
 #ifndef __DEFAULTSETTINGS_H__
 #define __DEFAULTSETTINGS_H__
 // clang-format off
+//
```
cquesnel, line 4: Why the extra `//` ?

cquesnel (changes requested):

cquesnel (approved):

PR #25, opened 2021-02-10
Feat/swarinfo 153 stream abstraction

cquesnel:


src/Main.cpp
```diff
@@ -75,26 +101,60 @@ class HostUartCommTask : public AbstractTask<2 * configMINIMAL_STACK_SIZE> {
...
+
+        uint8_t buffer[5];
+        buffer[sizeof(buffer) - 1] = 0;
+
+        // Wait for connection
+        while (true) {
+            if (std::optional<TCPClientWrapper> socket = SocketContainer::getHostClientSocket()) {
+                while (true) {
+                    auto ret = socket.value().receive(buffer, sizeof(buffer) - 1);
+                    (void)ret;
```
t-constantin, line 133: what is this line doing? (for my understanding)

src/bsp/include/bsp/IHostUart.h
```diff
@@ -1,12 +1,14 @@
 #ifndef __IHOSTUART_H__
 #define __IHOSTUART_H__
 
+#include <common/IProtobufStream.h>
 #include <cstdint>
 
-#define HOST_UART_MAX_MESSAGE_LENGTH UINT16_MAX
+#define HOST_UART_MAX_MESSAGE_LENGTH 1024
```
t-constantin, line 7: Should we document these lengths for when this is used accross platforms? (such as android app)

src/bsp/src/posix/CMakeLists.txt
```diff
@@ -10,18 +10,17 @@ set(LIB_SOURCES
     src/TCPClient.cpp
     src/SocketContainer.cpp
     src/SocketFactory.cpp
-    src/BSPUtils.cpp
     src/TCPUartMock.cpp
     src/SettingsContainer.cpp
-)
+        )
```
t-constantin, line 15: indentation?

src/bsp/src/posix/include/TCPUartMock.h
```diff
@@ -2,11 +2,14 @@
 #define __TCPUARTMOCK_H__
 
 #include "bsp/IHostUart.h"
+#include <condition_variable>
```
t-constantin, line 5: I don't see those new headers used in the file. Still relevant?

src/bsp/src/stm32/src/HostUart.cpp
```diff
@@ -62,11 +70,22 @@ bool HostUart::send(const uint8_t* buffer, uint16_t length) {
     return ret;
 }
 
-int32_t HostUart::receive(uint8_t* buffer, uint16_t length) const {
-    (void)buffer;
-    (void)length;
-    // TODO: Implement
-    return 0;
+bool HostUart::receive(uint8_t* buffer, uint16_t length) {
+    if (buffer == NULL || length > m_streamMemory.size()) {
```
t-constantin, line 74: isn't better to compare with nullptr? I believe NULL means 0

t-constantin (changes requested):
Nice work! Only small comments, but maybe someone else should validate more the logic.

src/Main.cpp
```diff
@@ -65,6 +64,33 @@ class BittyBuzzTask : public AbstractTask<6 * configMINIMAL_STACK_SIZE> {
...
+        AbstractTask(taskName, priority), m_logger(LoggerContainer::getLogger()) {}
+
+    ~UARTReadTask() override = default;
+
+  private:
+    ILogger& m_logger;
+    void task() override {
+
+        uint8_t buffer[5];
+        buffer[sizeof(buffer) - 1] = 0;
```
xgroleau, line 79: Shouldn't you null terminate after receiving? If it's for a string null termination, this way would be more verbose
`buffer[sizeof(buffer)-1] = '\0'`

src/Main.cpp
```diff
@@ -75,26 +101,60 @@ class HostUartCommTask : public AbstractTask<2 * configMINIMAL_STACK_SIZE> {
   private:
     void task() override {
         IHostUart& hostUart = BSPContainer::getHostUart();
+        uint8_t buffer[5];
+        buffer[sizeof(buffer) - 1] = 0;
```
xgroleau, line 105: Unused

src/Main.cpp
xgroleau: @t-constantin  It's simly to remove the compiler warning about unused variable,  we have a LOT of compiler checks on the hivemind.
Though removing the whole ret variable would fix this problem (thus ignoring the return) and remove the need of the explicit discard
`socket.value().receive(buffer, sizeof(buffer) - 1);`

src/Main.cpp
```diff
@@ -75,26 +101,60 @@ class HostUartCommTask : public AbstractTask<2 * configMINIMAL_STACK_SIZE> {
...
   public:
-    HostTCPCommTask(const char* taskName, UBaseType_t priority) :
-        AbstractTask(taskName, priority) {}
+    HostTCPCommTask(const char* taskName, UBaseType_t priority, TCPReadDemoTask& tcpReadTask) :
+        AbstractTask(taskName, priority), m_tcpReadDemoTask(tcpReadTask) {}
 
     ~HostTCPCommTask() override = default;
 
   private:
+    TCPReadDemoTask& m_tcpReadDemoTask;
```
xgroleau, line 151: So the HostTCPCommTask is responsible for the tcpReadDemoTask? could it create it on construction instead of injecting it if it's responsible of it? IDK if it would make more sense

src/bsp/include/bsp/IHostUart.h
```diff
@@ -1,12 +1,14 @@
...
+#define HOST_UART_STREAM_SIZE 4096
```
xgroleau, line 8: Should we prefer constexpr over defines? More of a general comment on our coding standards

src/bsp/src/posix/src/SocketContainer.cpp
```diff
@@ -3,6 +3,7 @@
 #include "TCPClient.h"
 #include "bsp/SettingsContainer.h"
 #include "logger/LoggerContainer.h"
+#include <optional>
```
xgroleau, line 6: Unused header

src/bsp/src/posix/src/SocketFactory.cpp
```diff
@@ -3,6 +3,7 @@
 #include <arpa/inet.h>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <optional>
```
xgroleau, line 6: unused header

src/bsp/src/posix/src/TCPUartMock.cpp
```diff
@@ -2,14 +2,17 @@
...
 #include <ros/ros.h>
 #include <task.h>
 
 void TCPUartMock_listenTask(void* param) {
     auto* test = static_cast<TCPUartMock*>(param);
     test->waitForClient();
+
+    while (true) {
+        vTaskDelay(1000);
+    }
```
xgroleau, line 15: why do we keep the task alive?

src/bsp/src/posix/src/UserInterface.cpp
```diff
@@ -17,6 +18,9 @@ int UserInterface::print(const char* format, va_list args) const {
 
     int retValue = vsnprintf(buffer, bufferSize, format, args);
 
+    vPortEnterCritical();
     ROS_INFO("%s", buffer);
+    vPortExitCritical();
```
xgroleau, line 23: Was this for debug or ROS_INFO actually needs to be critical?

src/Main.cpp
cquesnel: it was to prevent an error being thrown because of an unused parameter (because I needed to debug the value) but it is not needed anymore

src/bsp/src/stm32/src/HostUart.cpp
```diff
@@ -62,11 +70,22 @@ bool HostUart::send(const uint8_t* buffer, uint16_t length) {
...
+        m_logger.log(LogLevel::Warn, "Invalid parameters for UART receive");
+        return false;
+    }
+
+    m_receivingTaskHandle = xTaskGetCurrentTaskHandle();
+    while (CircularBuff_getLength(&m_stream) < length) {
+        // Gets notified everytime a new packet is appended to stream
+        ulTaskNotifyTake(pdTRUE, 500);
+    }
+    m_receivingTaskHandle = NULL;
```
xgroleau, line 84: Why change the taskhandle everytime? in case the HostUart receives in different thread?

src/bsp/include/bsp/IHostUart.h
cquesnel: Added to confluence

xgroleau (changes requested):
Clean, the syscall wrapper made me **ahah**.
Small comments and questions as usual

src/bsp/src/stm32/src/HostUart.cpp
cquesnel: nullptr is also equal to 0, but will change because we are in c++

src/Main.cpp
cquesnel: Changed for '\0', but honestly most of the code in the main will disappear pretty soon.

src/Main.cpp
cquesnel: All of this is gonna be removed the moment we have a dispatcher. Does it really give anything to make it clean now?

src/bsp/src/posix/src/TCPUartMock.cpp
cquesnel: Because the kernel sometimes stops working if we don't hahaha
Will add a comment

src/bsp/src/posix/src/UserInterface.cpp
cquesnel: Making it critical prevents a context switch (by disabling signals) in the middle of the function. I'd rather leave it there and be safe. FreeRTOS documentation does give printf as an example of a function that should be surrounded with disabled signals

src/bsp/src/stm32/src/HostUart.cpp
cquesnel: Exactly

src/bsp/include/bsp/IHostUart.h
cquesnel: I don't really have an opinion on that

src/Main.cpp
t-constantin: I see, thanks

t-constantin (approved):

src/bsp/src/posix/src/TCPUartMock.cpp
xgroleau: F

xgroleau (approved):
LGTM 👍

src/bsp/src/posix/include/BSPUtils.h
```diff
@@ -1,19 +1,53 @@
...
+
+            retVal = std::invoke(func, args...);
+            isFinished = true;
+        });
+        t.detach();
+
+        vPortExitCritical();
+
+        while (!isFinished) {
+            vTaskDelay(10);
```
stlaperriere, line 45: A great piece of art that is

stlaperriere (approved):

PR #26, opened 2021-02-11
Feat/swarinfo 197 dispatch message

xgroleau:
Added queue, message handling and writng to destination

 we may move the thread safe queue in a distinc lib, though it,s there for now


src/message-handler/include/message-handler/IMessageSender.h
```diff
@@ -0,0 +1,16 @@
...
+#define __IMESSAGESENDER_H_
+
+/**
+ *@brief A class to pipe messages to a stream and serialize them */
+class IMessageSender {
+  public:
+    virtual ~IMessageSender() = default;
+
+    /**
+     *@brief process an item in the queue, serializes it and sends it to the deserializer
```
cquesnel, line 11: Serializes it and then deserializes it? Don't you mean sends it to the stream?

src/message-handler/src/MessageDispatcher.cpp
```diff
@@ -0,0 +1,98 @@
...
+    if (m_deserializer.deserializeFromStream(message)) {
+
+        uint32_t destinationId = message.getDestinationId();
+        if (destinationId == m_bspUuid) {
+            return dispatchMessage(message);
+        }
+
+        // Message is for a remote host
+        m_logger.log(LogLevel::Debug, "Message sent to remote queue");
+        m_remoteOutputQueue.push(message);
```
cquesnel, line 28: I don't remember if the queue does a copy, but if it does not, message will go out of scope

src/message-handler/src/MessageDispatcher.cpp
```diff
@@ -0,0 +1,98 @@
...
+        return true;
+    }
+
+    // Could not find the destination
+    m_logger.log(LogLevel::Warn, "Destination could not be found");
```
cquesnel, line 33: If we end up here, it's because `deserializeFromStream()` failed, not because destination could not be found

src/message-handler/src/MessageHandlerContainer.cpp
```diff
@@ -0,0 +1,23 @@
+#include "MessageHandlerContainer.h"
+#include <cpp-common/CircularQueueStack.h>
+
+ThreadSafeQueue<MessageDTO>& MessageHandlerContainer::getBuzzMsgQueue() {
+    static CircularQueueStack<MessageDTO, 8> s_buzzMsgQueue;
```
cquesnel, line 5: should we maybe move the queue sizes to a macro or a const?

cquesnel: Or even maybe as a compile setting?

cquesnel (changes requested):

src/message-handler/src/MessageDispatcher.cpp
xgroleau: It creates a copy

src/message-handler/src/MessageHandlerContainer.cpp
xgroleau: I'll put a const expression for now

src/message-handler/src/MessageDispatcher.cpp
cquesnel: good

cquesnel (approved):

PR #27, opened 2021-02-11
Remove FreeRTOS from linux

cquesnel:


src/bsp/src/posix/src/BSP.cpp
```diff
@@ -38,13 +34,14 @@ void exampleTopicPublish(void* param) {
...
-        vTaskDelay(loopRate);
+        Task::delay(loopRate);
     }
 }
 
 BSP::BSP() :
     m_rosWatchTask("ros_watch", tskIDLE_PRIORITY + 1, rosWatcher, NULL),
-    m_exampleTopicPublishTask("ros_watch", tskIDLE_PRIORITY + 1, rosWatcher, &m_rosNodeHandle) {}
+    m_exampleTopicPublishTask(
+        "ros_watch", tskIDLE_PRIORITY + 1, exampleTopicPublish, &m_rosNodeHandle) {}
```
xgroleau, line 44: Why did the function pointer changed here?

xgroleau (approved):
LGTM 👍, though there is a function pointer that changed, and don't forget to bump propolis before merging

Did you remove the syscall wrapper? I didn't see the file while reviewing.

src/bsp/src/posix/src/BSP.cpp
cquesnel: Because it was starting the same function (`rosWatcher`) twice. An error introduced when you refactored to AbstractTasks

cquesnel (commented):
And yes, I removed BSPUtils, so the wrapper isn't used anywhere

PR #28, opened 2021-02-14
Feat/swarinfo 200 integration for demo

cquesnel:


src/Main.cpp
```diff
@@ -86,26 +93,28 @@ class TCPMessageDispatcher : public AbstractTask<2 * configMINIMAL_STACK_SIZE> {
...
-            while (true) {
-                if (!messageDispatcher.deserializeAndDispatch()) {
-                    m_logger.log(LogLevel::Warn, "Fail to deserialize/dispatch TCP");
+        while (true) {
+            auto socket = SocketContainer::getHostClientSocket();
+            if (socket) {
+                HiveMindHostDeserializer deserializer(socket.value());
+                MessageDispatcher messageDispatcher(MessageHandlerContainer::getBuzzMsgQueue(),
+                                                    MessageHandlerContainer::getHostMsgQueue(),
+                                                    MessageHandlerContainer::getRemoteMsgQueue(),
```
g-lesssard, line 102: Who's the remote? Cellphone and/or ESP?

src/bsp/src/posix/src/SocketContainer.cpp
```diff
@@ -3,12 +3,15 @@
 #include "TCPClient.h"
 #include "bsp/SettingsContainer.h"
 #include "logger/LoggerContainer.h"
+#include "LockGuard.h"
 
 std::optional<TCPClientWrapper> SocketContainer::getHostClientSocket() {
 
     static std::optional<TCPClient> s_clientSocket = {};
     static std::optional<TCPClientWrapper> s_wrapper = {};
+    static Mutex s_mutex = Mutex(500);
```
g-lesssard, line 12: I guess 500 is the delay to acquire the mutex with freertos. Could we reduce this delay? 500 seems a bit long and at this length, it should be either blocking or a lower timeout.

g-lesssard (approved):
GG WP

src/bsp/src/posix/src/SocketContainer.cpp
cquesnel: I left a longer delay because the mutex locks the socket ::connect which is a long operation

src/Main.cpp
cquesnel: Remote is the ESP, but in this case, we are treating the cellphone (so the uart communication) as remote for the sake of the demo

PR #31, opened 2021-02-16
Feat/swarinfo 188 add typechecking buzz

xgroleau:
Added type checking for bittybuzz

I did not add nativefunction and userdata since they don't seem to be used/available in the VM


src/bittybuzz/include/bittybuzz/BittyBuzzUserFunctions.h
```diff
@@ -12,18 +12,65 @@ namespace BittyBuzzUserFunctions {
...
     void logInt();
 
     /**
      *@brief Logs a string value
-     *
      *@details This closure expects one stringId parameter, the string to log. Returns nothing
      * */
     void logString();
 
+    /**
```
stlaperriere, line 25: The `BittyBuzzUserFunctions` namespace contains some user-defined functions right? Does that mean that type checking will have to be managed and implemented by the user or is this feature a part of the "platform"?

stlaperriere (approved):
Looks fine to me! Although I'm still a complete noob with everything related to Buzz.

src/bittybuzz/include/bittybuzz/BittyBuzzUserFunctions.h
xgroleau: UserFunctions are the function that we make available to the user. In other words, we add functions to the buzz VM, this is those function and the user can use them in their buzz script.

PR #34, opened 2021-02-27
Feat/swarinfo 220 callbackregistration

xgroleau:
Can now register closures and those closures will be called on a functioncallrequest message


src/bittybuzz/src/BittyBuzzClosureRegister.cpp
```diff
@@ -0,0 +1,43 @@
+#include "BittyBuzzClosureRegister.h"
+#include <bbzvm.h>
+#include <cstring>
+
+BittyBuzzClosureRegister::BittyBuzzClosureRegister() = default;
+
+bool BittyBuzzClosureRegister::registerClosure(const char* functionName,
+                                               bbzheap_idx_t closureHeapIdx) {
+    if (m_closureRegistersLength >= m_maxSize) {
```
g-lesssard, line 9: What happens in the case that the m_closureRegisterLength has exceeded its maxlength? Do those values directly correspond to the size of the heap used? Is it corruputed or is there some protections internally to the bbvm?

Similar to the previos comment on the interface of this class, the allocation could be done internally to avoid this sort of weird state.

g-lesssard (changes requested):
Good work! Just some clarification needed on the level of reponsability of the interface used to register callabacks and some minor typos

src/bittybuzz/include/bittybuzz/IBittyBuzzClosureRegister.h
```diff
@@ -0,0 +1,29 @@
...
+    virtual ~IBittyBuzzClosureRegister() = default;
+
+    /**
+     *@brief register un function to the list
+     *@param functionName the name of the function
+     *@param closureHeapIdx a pointer to the heap to the closure. When registering, the function
+     *will be made permanent.
+     *@return true on success, false if not (i.e. no more space in the list, or the heapidx is not a
+     *closure)*/
+    virtual bool registerClosure(const char* functionName, bbzheap_idx_t closureHeapIdx) = 0;
```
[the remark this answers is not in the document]
xgroleau: We provide the heap address where the closure is already stored, the function is defined/allocated by buzz in the script. (so the function doesn't do any malloc, it just disables the GC on the object). Check the BittyBuzzUserFunctions::registerClosure() on how it's used. The address is the addressed passed by the function the the buzzscript.


The tests are really juste to mock, they aren't valid closures.  The 42 value on the new isn't even valid and I know that in the tests the closure won't be called. If the closure is called, we would have undefined behavior here probably. It's just to have a valid closure object (so the typecheck passes). The 42 is supposed to be an intptr_t (I guess a pointer to the function in the bzz code)  but it isn't  valid in this case.

src/bittybuzz/src/BittyBuzzClosureRegister.cpp
xgroleau: I'm not sure what you mean, here it's really just a list/array of addresses in the heap associated with a string. It shouldn't exceed since there is a check at this line. the m_maxSize is the max size of the array. There is no allocation here, It's already allocated by the buzz script, I just make it permanent on line 13 so the GC doesn't destroy it (so the closure lives during the whole program)

g-lesssard (approved):

PR #35, opened 2021-03-02
Feat/swarinfo 156 call host functions

xgroleau:
Can now call function from buzz and added random number generate, on the STM32 it uses the hardware RNG


src/bittybuzz/tests/BittyBuzzMessageServiceUnitTests.cpp
```diff
@@ -0,0 +1,224 @@
...
+    LoggerInterfaceMock* m_loggerMock;
+    BSPInterfaceMock* m_bspMock;
+    int m_logCounter = 0;
+    const uint16_t m_bspUUID = 42;
+    const uint16_t m_remoteUUID = 43;
+    const int64_t m_argVal = 42;
+    const std::string m_functionName = "print";
+    std::string m_logLastFormat;
+
+    // DTO
```
g-lesssard, line 23: 3 accolades?

src/bsp/src/stm32/hal/stm32f429zi/stm32f429zi.ioc
```diff
@@ -99,7 +99,7 @@ ProjectManager.ProjectFileName=stm32f429zi.ioc
 RCC.ADC12outputFreq_Value=72000000
 FREERTOS.Tasks01=defaultTask,24,128,StartDefaultTask,Default,NULL,Dynamic,NULL,NULL
 PD9.GPIO_Label=STLK_TX [STM32F103CBT6_PA2]
-Mcu.PinsNb=41
```
g-lesssard, line 102: We gained a pin?

g-lesssard (approved):
LGTM. Cannot validate the random number generator, but it looks good. Not entirely sure why we need it but, its there.

src/bsp/src/stm32/hal/stm32f429zi/stm32f429zi.ioc
xgroleau: That is a very good question lol

stlaperriere (approved):
Nothing much to say except good job!

PR #37, opened 2021-03-03
Feat/swarinfo 264 integrate usb driver

cquesnel:
All C code is generated. Do not review!


cmake/usb/FindUSB.cmake
```diff
@@ -0,0 +1,115 @@
+if(NOT USB_FIND_COMPONENTS)
```
xgroleau, line 1: Should the whole file go in stm32/FindUSB.cmake ,like the FindBSP, FindHAL, FindCMSIS? Or if we have another chip we will have both port in the same file?

src/Main.cpp
```diff
@@ -219,7 +219,7 @@ int main(int argc, char** argv) {
     s_tcpDispatchTask.start();
     s_uartMessageSender.start();
     s_tcpMessageSender.start();
-    s_spiMessageSender.start();
+    // s_spiMessageSender.start();
```
xgroleau, line 222: Forgot to uncomment

xgroleau (approved):
LGTM 👍

src/Main.cpp
cquesnel: No it breaks if we don't have an ESP plugged in. So until we all work with a HiveSight, we will leave commented

cmake/usb/FindUSB.cmake
cquesnel: It could go there!

PR #38, opened 2021-03-03
Feat/swarinfo 259 request

xgroleau:
Added HiveMindAPI, for the user, will always stay internal to the hivemind and it's host. And the SwarmAPI for the ESP side, shouldn't reach the hivemind. Though the SwarmAPI may eventually go to another distinct library

Note: I added a greeting in the Main, though for the Uart I think we wait for it to sends bytes, though if it waits for us to greet, we will have a problem.


src/bittybuzz/src/BittyBuzzMessageHandler.cpp
```diff
@@ -133,11 +141,17 @@ bool BittyBuzzMessageHandler::handleResponse(const ResponseDTO& response) {
     if (const auto* gResp = std::get_if<GenericResponseDTO>(&variantResp)) {
         return handleGenericResponse(*gResp);
     }
+    if (std::holds_alternative<HiveMindApiResponseDTO>(variantResp)) {
+        m_logger.log(LogLevel::Warn, "Receivec Hivemind Resp in buzz queue");
```
cquesnel, line 145: Reveivec

src/message-handler/src/HiveMindApiRequestHandler.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+HiveMindApiRequestHandler::HiveMindApiRequestHandler(const IBSP& bsp, ILogger& logger) :
+    m_bsp(bsp), m_logger(logger) {}
+
+HiveMindApiResponseDTO HiveMindApiRequestHandler::handleRequest(
+    const HiveMindApiRequestDTO& request) {
+
+    const std::variant<std::monostate, IdRequestDTO>& vReq = request.getRequest();
+    if (std::holds_alternative<IdRequestDTO>(vReq)) {
+        return HiveMindApiResponseDTO(IdResponseDTO(m_bsp.getUUId()));
```
cquesnel, line 11: Same comment as in Propolis. Difference between a `uint16_t` and a `uint32_t`

src/Main.cpp
```diff
@@ -166,8 +169,9 @@ class TCPMessageSender : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
             if (socket) {
                 HiveMindHostSerializer serializer(socket.value());
                 MessageSender messageSender(MessageHandlerContainer::getHostMsgQueue(), serializer,
-                                            m_logger);
+                                            BSPContainer::getBSP(), m_logger);
 
+                messageSender.greet();
```
cquesnel, line 174: I think the greeting process should be a little more complex than just sending a message. At least, we should wait for a specific response. Otherwise let's say we are connected by USB but on a device which has not yet launched the app. We send a greet and after that, we expect to be connected while we aren't really. I would maybe have a loop that greets at a specific rate (once we now we are physically connected, this will be possible with USB) until we get a confirmation. We should also have a state somewhere saying on which port a successful greeting has been achieved. (Later this could light a LED up).

This is kind of a big change so it could be done in another PR.

cquesnel (changes requested):

src/Main.cpp
xgroleau: Would this handshake go in the driver? Because on the TCP side I know that if I am connected, someone is listening on the other side. So the driver would juste send a specific byte while waiting and once it received one, it knows the connection is established like TCP.

cquesnel: Well it could go further than just knowing if someone else is on the other end of the line. The handshake could include a version check to be sure messages are compatible between the two entities. I just think it is a good idea to know if we are correctly understood on the other side before trying more messages.

@stlaperriere @g-lesssard What do you guys think?

xgroleau: No need for version check though with protobuf, if the fields number id corresponds, they can mismatch and protobuf will only deserialize what it knows

xgroleau: But I see what you mean, maybe another PR to add a greetingresponse

g-lesssard: I agree with the loop mechanism proposed, to make sure we have a connection with the module we want to greet. That is the purpose of a handshake. However, since this is more specific logic, it shouldn't be handled within Propolis but by the HiveMind and HiveConnect firmwares (the HiveMind will greet the ROS-bridge and expect a response and the ESP will greet the HiveMind and expect a repsonse). As you said, a version field could be added in this handshake to detect a breaking change between versions, but this might be more useful in the future than now, as we are currently adding messages more than modifying existing ones.

cquesnel: Good, so I will approve and we can create a task to make the handshake loop later on

cquesnel (approved):

src/Main.cpp
stlaperriere: I agree with you guys

PR #40, opened 2021-03-04
Feat/swarinfo 233 deca driver for two decas

cquesnel:
Add a function to switch between decawaves for spi communication


src/bsp/src/stm32/hal/decawave/platform/stm32f429zi/include/deca_port.h
```diff
@@ -34,6 +40,24 @@ void deca_setFastRate();
  */
 void deca_init();
 
+/**
+ * @brief Choose which decawave is selected for the next SPI communication
+ * @param selectedDevice The device to select
+ */
+void deca_selectDevice(decaDevice_t selectedDevice);
```
xgroleau, line 47: Wouldn't it be easier to have two getter for a static struct variable. One getter for the decaA and another to the decaB. Then pass that struct around. The struct would contain the necessary info, so the NssPort and Pin in this case. Avoiding to call Select device each time and just get the struct directly?

xgroleau (changes requested):
Looks good, just a small question, could we use two static structs instead of switching the selected device?

g-lesssard (approved):

src/bsp/src/stm32/hal/decawave/platform/stm32f429zi/src/deca_port.c
```diff
@@ -2,30 +2,52 @@
...
-    GPIO_InitStruct.Pull = GPIO_NOPULL;
-    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_HIGH;
-    HAL_GPIO_Init(DW_RESET_A_GPIO_Port, &GPIO_InitStruct);
-    HAL_GPIO_WritePin(DW_RESET_A_GPIO_Port, DW_RESET_A_Pin, GPIO_PIN_SET);
+    gpioInitStruct.Pin = selectedDecaResetPin;
+    gpioInitStruct.Mode = GPIO_MODE_OUTPUT_OD;
+    gpioInitStruct.Pull = GPIO_NOPULL;
+    gpioInitStruct.Speed = GPIO_SPEED_FREQ_HIGH;
+    HAL_GPIO_Init(selectedDecaResetPort, &gpioInitStruct);
+    HAL_GPIO_WritePin(selectedDecaResetPort, selectedDecaResetPin, GPIO_PIN_SET);
```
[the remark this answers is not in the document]
cquesnel: No the problem is, we have to put the pin in different modes throughout the reset process. It's a requirement from decawave as once it has been reset, the pin becomes an output of the chip so on the ST side, we can't just pull it to a value anymore

src/bsp/src/stm32/hal/decawave/platform/stm32f429zi/include/deca_port.h
cquesnel: Good idea, changed it for a structure pointer that is passed around

xgroleau (approved):

PR #46, opened 2021-03-16
Feat/swarinfo 300 handle disconnections on uart mock

g-lesssard:
A simple fix that blocks on disconnection until the connection has been restored. The greet functionality is not supported yet, as it needs some rework. Also, added a default port for server socket to simplify debugging.


g-lesssard:
Changes incorporated in other PR

PR #47, opened 2021-03-16
Feat/swarinfo 256 mocking spi interface

g-lesssard:
Added the interface for esp as a proper stream in the message handler. For now, the code is duplicated for the mock from the uart. Since the two are still under development, they do not have the same interface but in the future, they could share one so that the code is not duplicated and we would have two instances of the same mock class. for these serial links. The same goes if we switch to the usb connection.

Also added handling disconnection on the mock for the uart and spi classes


src/message-handler/include/message-handler/MessageHandlerContainer.h
```diff
@@ -36,6 +36,13 @@ namespace MessageHandlerContainer {
      *@brief get the remote message queue
      *@return A reference to the remote message queue */
     ThreadSafeQueue<MessageDTO>& getRemoteMsgQueue();
+
+    /**
+     *
+     *@brief get the network message queue
+     *@return A reference to the network message
+     */
+    ThreadSafeQueue<MessageDTO>& getNetworkMessageQueue();
```
xgroleau, line 45: No need for this queue, the RemoteMsgQueue was for the ESP, putting it on the Uart stream was purely temprorary for the old demo

src/Main.cpp
```diff
@@ -195,17 +195,54 @@ class SPIMessageSender : public AbstractTask<5 * configMINIMAL_STACK_SIZE> {
...
+
         auto& spi = BSPContainer::getSpiEsp();
         while (true) {
-            if (!spi.isBusy()) {
-                const char message[] = "Sending message to ESP from task";
-                spi.send((uint8_t*)message, sizeof(message));
+            if (spi.isConnected()) {
+
+                HiveMindHostSerializer serializer(spi);
+                MessageSender messageSender(MessageHandlerContainer::getNetworkMessageQueue(),
```
xgroleau, line 204: Use remotequeue, it was made for the ESP though the uart uses for the demo, I guess we could remove the uart task for now.

src/bsp/src/stm32/src/SpiEsp.cpp
```diff
@@ -64,6 +64,8 @@ bool SpiEsp::send(const uint8_t* buffer, uint16_t length) {
 
 bool SpiEsp::isBusy() const { return m_isBusy; }
 
+bool SpiEsp::isConnected() const { return true; }
```
xgroleau, line 67: Will this have actual logic eventually? If so maybe add a todo.

src/bsp/src/posix/src/SpiEspMock.cpp
```diff
@@ -0,0 +1,112 @@
+#include "SpiEspMock.h"
+#include <BaseTask.h>
+#include <Task.h>
+#include <logger/Logger.h>
+#include <ros/ros.h>
+
+void SpiMock_listenTask(void* param) {
+    auto* test = static_cast<SpiEspMock*>(param);
```
xgroleau, line 8: rename variable

xgroleau (changes requested):
Gj, pretty simple, small thing, remove the NetworkMessageQueue and use the RemoteMsgQueue (you may need to disable the uart thread for now) or you will have issues down the line since this queue is not handled in the dispatcher

src/message-handler/include/message-handler/MessageHandlerContainer.h
g-lesssard: I expect it will break the demo we have working now though?

xgroleau: Yes this will break the demo though since no more data will be read from the uart. But the uart has bugs anyway in it I think. Though once your code is working in simulation at the end of the sprint, we should be able to send a message via Phone->HM->ESP->ESP->HM->Robot.

xgroleau (approved):

PR #48, opened 2021-03-19
Feat/swarinfo 286 messages buzz stigmergy

xgroleau:
Now sending buzz messages

### DIFF
--- a/src/bittybuzz/src/BittyBuzzMessageHandler.cpp
+++ b/src/bittybuzz/src/BittyBuzzMessageHandler.cpp
@@ -1,6 +1,7 @@
 #include "BittyBuzzMessageHandler.h"
 #include "BittyBuzzSystem.h"
 #include "bbzvm.h"
+#include <bbzinmsg.h>
 #include <bsp/SettingsContainer.h>
 
 BittyBuzzMessageHandler::BittyBuzzMessageHandler(const IBittyBuzzClosureRegister& closureRegister,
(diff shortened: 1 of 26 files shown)


src/bittybuzz/src/BittyBuzzMessageHandler.cpp
```diff
@@ -180,6 +181,19 @@ bool BittyBuzzMessageHandler::handleGenericResponse(const GenericResponseDTO& re
     return true;
 }
 
+bool BittyBuzzMessageHandler::handleBuzzMessage(const BuzzMessageDTO& msg) {
+    (void)msg;
+    bbzmsg_payload_t payload;
+    payload.buffer = const_cast<uint8_t*>(msg.getPayload().data());
```
cquesnel, line 187: If we don't copy the message content, is there a risk the pointer value changes before the BBVM has time to ingest the message?

src/os/freertos/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ find_package(FreeRTOS COMPONENTS
             NONE
             REQUIRED)
 
+get_target_property(PROPOLIS_OS_COMMON_INCLUDES SwarmUS::Propolis::OS::Common INTERFACE_INCLUDE_DIRECTORIES)
```
cquesnel, line 28: Careful, you're overriding the PR fix we did before

src/os/posix/CMakeLists.txt
```diff
@@ -12,16 +12,19 @@ add_library(${LIB_ALIAS} ALIAS ${LIB_NAME})
 
 # Wraps the interface library in a static library
 get_target_property(PROPOLIS_OS_INCLUDES SwarmUS::Propolis::OS::POSIX INTERFACE_INCLUDE_DIRECTORIES)
+get_target_property(PROPOLIS_OS_COMMON_INCLUDES SwarmUS::Propolis::OS::Common INTERFACE_INCLUDE_DIRECTORIES)
```
cquesnel, line 15: See previous comment

cquesnel (changes requested):

src/bittybuzz/src/BittyBuzzMessageHandler.cpp
xgroleau: The bbzinmsg_queue_append deserializes it, so it copies the data and doesn't just keep a pointer. If we would copy it in a bbzmsg_paylaod_t it would be a useless copy, so we save a couple cpu cycles with the const cast

src/os/freertos/CMakeLists.txt
xgroleau: good catch, thanks for pointing it

cquesnel (approved):

PR #52, opened 2021-03-19
Feat/swarinfo 281 dw class

cquesnel:
Added a C++ driver to control a DW1000.
Example of the driver is given in InterlocManager where one decawave sends a message to the other on the same HiveSight
Renamed the CRC class because of a conflict with a define in the decawave driver


src/bsp/src/stm32/include/Decawave.h
```diff
@@ -0,0 +1,163 @@
...
+  public:
+    explicit Decawave(decaDevice_t spiDevice);
+    Decawave(decaDevice_t spiDevice, uint8_t channel, UWBSpeed speed);
+    ~Decawave() = default;
+
+    /**
+     * @brief Tries communicating by SPI with the DW1000 and configures ISR, GPIO, etc.
+     * @return true if successful, false otherwise.
+     */
+    bool start();
```
dubh3401, line 26: does it start anything? if it is only a config/init maybe name it that way

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -0,0 +1,307 @@
...
+    m_speed = speed;
+    configureDW();
+}
+
+void Decawave::setLed(DW_LED led, bool enabled) {
+    uint8_t dwGPIO = 0;
+    uint8_t dwGPIOValue = 0;
+
+    switch (led) {
+    case DW_LED::LED_0:
```
dubh3401, line 110: would'nt be simpler to have LED_x = DWT_GxPx and remove the switch/case?

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    if (!transmitInternal(buf, length, DWT_START_TX_IMMEDIATE | DWT_RESPONSE_EXPECTED)) {
+        return false;
+    }
+
+    receiveInternal(frame, 0, 0, true);
+    return true;
+}
+
+bool Decawave::transmitAndReceiveDelayed(uint8_t* buf,
```
dubh3401, line 246: confusing
transmitDelayedAndReceive

src/bsp/src/posix/include/InterlocManager.h
```diff
@@ -0,0 +1,14 @@
+#ifndef __INTERLOCMANAGER_H__
+#define __INTERLOCMANAGER_H__
+
+class InterlocManager : public IInterlocManager {
+  public:
+    InterlocManager() = default;
```
xgroleau, line 6: No need to declare it if it's default

src/bsp/include/bsp/BSPContainer.h
```diff
@@ -25,7 +26,7 @@ namespace BSPContainer {
     IHostUart& getHostUart();
 
     /**
-     * @brief Returns an instance of the CRC driver.
+     * @brief Returns an instance of the UserCRC driver.
```
xgroleau, line 29: UserCRC? it's the crc that is not used internally and that the user can use?

src/bsp/src/stm32/hal/decawave/platform/stm32f429zi/src/deca_mutex.c
```diff
@@ -5,11 +5,16 @@
 // saving mechanism is already handled by the OS. We can only enter/exit a critical code section.
 
 decaIrqStatus_t decamutexon(void) {
```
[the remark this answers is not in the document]
cquesnel: This function is defined by DW, I can't change it

src/bsp/src/stm32/hal/decawave/platform/stm32f429zi/src/deca_mutex.c
```diff
@@ -5,11 +5,16 @@
...
-    vPortEnterCritical();
+    if (!xPortIsInsideInterrupt()) {
+        vPortEnterCritical();
+    }
     return 0;
 }
 
 void decamutexoff(decaIrqStatus_t s) {
```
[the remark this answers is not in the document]
cquesnel: This function is defined by DW, I can't change it

src/bsp/src/stm32/hal/decawave/platform/stm32f429zi/src/deca_port.c
```diff
@@ -66,7 +69,7 @@ void deca_setSlowRate() {
 }
 
 void deca_setFastRate() {
-    DW_SPI->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_4;
+    DW_SPI->Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_32;
```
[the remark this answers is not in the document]
cquesnel: From my tests, it's the fastest stable speed I can get

src/bsp/src/stm32/include/Decawave.h
```diff
@@ -0,0 +1,163 @@
...
+     * @param enabled True if ON false if OFF
+     */
+    void setLed(DW_LED led, bool enabled);
+
+    /**
+     * @brief Sets the UWB channel to use
+     * @param channelNo Channel number between 1 and 7
+     * @return True if configured, false if invalid channel
+     */
+    bool setChannel(uint8_t channelNo);
```
xgroleau, line 40: Could we use an enum?

src/bsp/src/stm32/include/Decawave.h
```diff
@@ -0,0 +1,163 @@
...
+    bool transmitAndReceiveDelayed(uint8_t* buf,
+                                   uint16_t length,
+                                   uint64_t txTimestamp,
+                                   uint32_t rxAfterTxTimeUs,
+                                   UWBRxFrame& frame,
+                                   uint16_t rxTimeoutUs);
+
+    friend void rxCallback(const dwt_cb_data_t* callbackData, void* context);
+    friend void isrCallback(void* context);
+    friend void rxAsyncTask(void* context);
```
xgroleau, line 131: You could declare them static instead of friends, this would allow you to access private members I think even when you cast the context. And you could  probably set them private too

src/bsp/src/stm32/include/UserCRC.h
```diff
@@ -1,15 +1,15 @@
 #ifndef __CRC_H__
-#define __CRC_H__
+#define __USERCRC_H__
```
xgroleau, line 2: wrong ifndef here at line 1

src/bsp/src/stm32/src/BSP.cpp
```diff
@@ -1,100 +1,14 @@
...
-        Task::delay(100);
-        dwt_setgpiovalue(DWT_GxM2, 0);
-
-        Task::delay(100);
-        dwt_setgpiovalue(DWT_GxM3, 0);
-    }
-}
-
-BSP::BSP() : m_decaBlinkTask("deca_blink", tskIDLE_PRIORITY + 1, decaBlink, NULL) {}
+BSP::BSP() = default;
```
xgroleau, line 5: No need for constructor declaration if it's defaulted, you could remove it in the .h too

src/bsp/src/stm32/include/Decawave.h
```diff
@@ -0,0 +1,163 @@
...
+#include <functional>
+#include <memory>
+#include <task.h>
+
+enum class DW_LED { LED_0, LED_1, LED_2, LED_3 };
+enum class UWBSpeed { SPEED_110K, SPEED_850K, SPEED_6M8 };
+
+class Decawave {
+  public:
+    explicit Decawave(decaDevice_t spiDevice);
```
xgroleau, line 18: Why the explicit? Also why two constructor?

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -0,0 +1,307 @@
...
+    m_rxAsyncTask("dw_rx_task", tskIDLE_PRIORITY + 10, rxAsyncTask, this) {}
+
+Decawave::Decawave(decaDevice_t spiDevice, uint8_t channel, UWBSpeed speed) :
+    m_spiDevice(spiDevice),
+    m_speed(speed),
+    m_rxAsyncTask("dw_rx_task", tskIDLE_PRIORITY + 10, rxAsyncTask, this) {
+
+    if (channel > 0 && channel < 8) {
+        m_channelNo = channel;
+    }
```
xgroleau, line 49: What happens with a channelNo = -1? Could and enum class solve this issue?

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -0,0 +1,307 @@
...
+}
+
+bool Decawave::start() {
+    deca_selectDevice(m_spiDevice);
+    deca_setSlowRate();
+    uint32_t deviceID = 0;
+    uint8_t i = 0;
+    // Retry to read the deviceID 10 times before abandoning
+    while (deviceID != DWT_DEVICE_ID && i++ < 10) {
```
xgroleau, line 58: Do while would probably be legitimate here, also your i will be incremented once before doing the loop once, so it's actually 9 times I think

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    setLed(DW_LED::LED_0, true);
+
+    m_rxAsyncTask.start();
+
+    return true;
+}
+
+bool Decawave::setChannel(uint8_t channelNo) {
+    if (channelNo < 1 || channelNo > 7) {
```
xgroleau, line 91: Could you use the same if check as of line 47?

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -0,0 +1,307 @@
...
+    if (length > UWB_MAX_LENGTH) {
+        return false;
+    }
+    deca_selectDevice(m_spiDevice);
+
+    memcpy(m_txBuffer.data(), buf, length);
+
+    // Send two bytes more than requested because of the DW auto-generated CRC16
+    dwt_writetxdata(length + 2, m_txBuffer.data(), 0);
+    dwt_writetxfctrl(length + 2, 0, 0);
```
xgroleau, line 211: Could you put the 2 in a define or constexpr, would make the code more auto documented

xgroleau (changes requested):
Idk the deca api so I didn't check the code very thoroughly, though it seems good. Somme comments/questions, pretty small details though GJ!

stlaperriere (commented):
Simply a question about the scalability of the code. At some point, we will have to support one or many decawaves. Right now support for exactly 2 chips is hard-coded, since we declare A and B. If the new HiveBoard design can support up to (6? 8?) theoretical DWs, we should maybe provide some functionalities to scale easily the number of chips that are managed.

Of course, that's a problem for a future PR, but I thought it was worth mentioning. Otherwise, GJ, this is clean.

src/bsp/src/stm32/include/Decawave.h
cquesnel: Changed for init

src/bsp/src/stm32/src/Decawave.cpp
cquesnel: It wasn't simpler because C++ class enums can't directly convert to their int value, but I added a utils function to do it. It's much cleaner. Also removed the conversion between our speed enum and the DW macros

src/bsp/include/bsp/BSPContainer.h
cquesnel: It's the CRC defined by the user (us) if you have a better name I can change it

src/bsp/src/stm32/include/Decawave.h
cquesnel: Changed for enum

src/bsp/src/stm32/include/Decawave.h
cquesnel: The explicit comes from a clang suggestion that single argument constructors should be marked explicit.
As for the second constructor, when I wrote the class I wasn't sure (and still am not as to how we would construct these objects ie configure them at runtime or through the constructor) so I left both options. It's probably going to end up being the second constructor with our default values and sometimes change them at runtime for tests

src/bsp/src/stm32/src/Decawave.cpp
cquesnel: Changed for enum class, but it was a uint8_t so it can't go lower than 0

src/bsp/src/stm32/src/Decawave.cpp
cquesnel: No need because of enum class

src/bsp/src/stm32/src/Decawave.cpp
cquesnel: Changed for do while

src/bsp/src/stm32/hal/decawave/platform/stm32f429zi/src/deca_port.c
```diff
@@ -1,9 +1,12 @@
...
+#include "deca_device_api.h"
 #include "main.h"
 #include "stm32f4xx_hal_conf.h"
 
-static decaNSSConfig_t g_decaNssConfigA = {DW_NSS_A_GPIO_Port, DW_NSS_A_Pin};
-static decaNSSConfig_t g_decaNssConfigB = {DW_NSS_B_GPIO_Port, DW_NSS_B_Pin};
+static decawaveDeviceConfig_t g_decaNssConfigA = {
+    DW_NSS_A_GPIO_Port, DW_NSS_A_Pin, DW_IRQn_A_GPIO_Port, DW_IRQn_A_Pin, 0, NULL, NULL};
+static decawaveDeviceConfig_t g_decaNssConfigB = {
+    DW_NSS_B_GPIO_Port, DW_NSS_B_Pin, DW_IRQn_B_GPIO_Port, DW_IRQn_B_Pin, 1, NULL, NULL};
```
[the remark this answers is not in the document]
cquesnel: The thing is there will be a maximum number of DWs based on the hardware layout. This part of the code will still need to exist to map each possible DW to its pins, we can't just magically make them appear. Also, using an enum vs just indexing the DWs with an int prevents (not as much as if it were a C++ class enum, but still more than an int) trying to access a DW that doesn't exist.

That being said, having all configs seperatly with a switch to return the good one, wasn't the best of idea's. I've change it for an array of configs where the index maps to the enum value of `DW_A`, `DW_B`, etc. The size of the array is scaled by `DWT_NUM_DW_DEV` defined in the DW API which also tells the API the number of DWs it should have in it's internal array.

In the user code, we will end up having some more dynamic way of choosing the devices, but as this is at hardware level, I don't think there's much more I can do. In the end, if we have 6 ports on the HiveBoard, there will be 6 entries in the array and it is in the user code that we will choose which DW to use based on which one responds when we try to communicate with it.

If you have another idea, feel free to share

src/bsp/include/bsp/BSPContainer.h
xgroleau: HCRC or HardwareCRC maybe?

src/bsp/src/stm32/include/Decawave.h
xgroleau: I personally prefer at the construction time, but it's fine keep both too

xgroleau (approved):
LGTM 👍

src/bsp/src/stm32/hal/decawave/platform/stm32f429zi/src/deca_port.c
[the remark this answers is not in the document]
stlaperriere: Ok this seems legit to me. So in the end, it will be the programmer's job to scale the number of DW devices according to what he has plugged on the board. That's absolutely fair four our use-case (if our devices were general-purpose with lots of different configs possible, I would have prefered that the config of the right number de peripherals happened automatically based on some data loaded in the EEPROM/flash on bootup, but we are not Trilliant, so this is absolutely fine 😆 ).

stlaperriere (approved):

PR #57, opened 2021-03-27
Feat/swarinfo 299 greeting and disconnections

xgroleau:
Added greeting and handling disconnections

- Wait for greet on connection
- Handles greet even when connection has been established
- When the stream disconnects, it retries to reconnect

Couple changes in the high level bsp architecture

- No more SPI mock or Uart mock or whatever, a ICommInterface was created, and then you just get the connected interface
- This made the need to remove a lot of files
- The TCPServer tries to reconnect in another thread (for the remote)

USB connection was not done since the isConnected is a bit problematic as of now


src/Main.cpp
```diff
@@ -324,25 +204,28 @@ int main(int argc, char** argv) {
...
-    static TCPMessageDispatcher s_tcpDispatchTask("tcp_dispatch", tskIDLE_PRIORITY + 1);
-    static SpiMessageDispatcher s_spiDispatchTask("spi_dispatch", tskIDLE_PRIORITY + 1);
-    static UartMessageSender s_uartMessageSender("uart_send", tskIDLE_PRIORITY + 1);
-    static TCPMessageSender s_tcpMessageSender("uart_send", tskIDLE_PRIORITY + 1);
-    static SPIMessageSender s_spiMessageSender("spi_send", tskIDLE_PRIORITY + 1);
-    static USBMessageSender s_usbMessageSender("usb_send", tskIDLE_PRIORITY + 5);
-    static USBMessageDispatcher s_usbMessageDispatcher("usb_receive", tskIDLE_PRIORITY + 5);
     static InterlocTask s_interlocTask("interloc", tskIDLE_PRIORITY + 5);
 
+    static MessageDispatcherTask s_hostDispatchTask("tcp_dispatch", tskIDLE_PRIORITY + 1, NULL,
```
t-constantin, line 209: Shouldn't we keep tsk priorities defined somewhere? instead of + X after idle

t-constantin (approved):
Code looks very clean!
I only have a small question about task priorities

g-lesssard (changes requested):
I like the refac done on the main, less tasks to create that are basically the same. A bit harder to understand the sequence at first but is cleaner.

Some issue with the TCP server, I suggested some changes to fix it.

For the greet, everything seems good I should be able to test it once the tcp server is fixed

stlaperriere (changes requested):
Good job, looks clean. Just a little question about the classes to manage greets, but its quite minor.

src/message-handler/include/message-handler/IGreetHandler.h
```diff
@@ -0,0 +1,15 @@
+#ifndef __IGREETHANDLER_H_
+#define __IGREETHANDLER_H_
+
+class IGreetHandler {
```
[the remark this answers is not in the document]
xgroleau: It was this way initially but I separated it since one is blocking and strictly using the serializer and deserializer, while the other function strictly uses the queue and is non-blocking. So the two functions did not share any members. 

Also since it has the serializer and deserializer, the message dispatcher threads would endup with a serializer. For exemple calling greet in the dispatcher thread would cause two tread to write on the stream (and the message sender writes on the stream too) causing a race condition. Seperating it in two classes removes this concern.

src/Main.cpp
xgroleau: Added define at the top of the main

src/bsp/src/posix/src/TCPServer.cpp
```diff
@@ -96,7 +96,10 @@ void TCPServer::waitForClient() {
     // Always tries to reconnect to client
     while (m_serverFd > 0) {
         if (!m_connected) {
-            ::close(m_clientFd.value());
+            // Close socket
+            if (m_clientFd) {
+                ::close(m_clientFd.value());
```
g-lesssard, line 101: Only call close if the m_clientFd has a value. You should clear the optional afterwards

g-lesssard (approved):
Approved, though missing optional check before closing in the TCP server

stlaperriere (approved):

PR #59, opened 2021-03-29
Added err handling and fixed subsequent greet

xgroleau:
Added some error handling and fixed subsequent greet not handled properly


g-lesssard (approved):
LGTM

PR #62, opened 2021-03-30
Fix/integration

cquesnel:
Working demo!

### DIFF
--- a/src/bsp/src/posix/src/TCPServer.cpp
+++ b/src/bsp/src/posix/src/TCPServer.cpp
@@ -82,10 +82,8 @@ bool TCPServer::receive(uint8_t* buffer, uint16_t length) {
 bool TCPServer::isConnected() const { return m_connected; }
 
 void TCPServer::close() {
-    m_connected = false;
-    if (m_clientFd) {
-        ::close(m_clientFd.value());
-    }
+    TCPServer::closeClient();
+
     if (m_serverFd > 0) {
         ::close(m_serverFd);
         m_serverFd = -1;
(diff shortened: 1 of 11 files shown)


src/bsp/src/stm32/src/TCPClient.cpp
```diff
@@ -1,7 +1,7 @@
 #include "TCPClient.h"
 
 TCPClient::TCPClient(int socket, sockaddr_in address, ILogger& logger) :
-    m_logger(logger), m_connected(socket > 0), m_socketFd(socket), m_address(address) {}
+    m_logger(logger), m_connected(socket >= 0), m_socketFd(socket), m_address(address) {}
```
g-lesssard, line 4: File descriptor 0 is stdin, why consider it connected for a socket?

src/bsp/src/posix/src/TCPServer.cpp
```diff
@@ -86,7 +86,7 @@ void TCPServer::close() {
     if (m_clientFd) {
         ::close(m_clientFd.value());
```
g-lesssard, line 87: Reset m_clientFd after closing. Also, could unify logic on check before closing

g-lesssard (approved):
Small comments

src/bsp/src/stm32/src/TCPClient.cpp
cquesnel: Because we aren't on UNIX and lwip doesn't use files. The return value of it's calls is nothing more than an index in an array. So the first started socket is at 0. Without this, the code just doesn't work.

src/bsp/src/posix/src/TCPServer.cpp
cquesnel: @xgroleau

src/bsp/src/stm32/src/TCPClient.cpp
g-lesssard: Good point

PR #63, opened 2021-03-31
Two way ranging and antenna delay calibration

dubh3401:
many TODO's and comment as markers for futur me


src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -1,7 +1,26 @@
...
 #include "bsp/BSPContainer.h"
 #include <Task.h>
+#define NB_CALIB_TRIES 10
 
+bool isFrameGood(UWBRxFrame frame, ILogger& logger) {
+
+    if (frame.m_status == UWBRxStatus::FINISHED) {
+        // logger.log(LogLevel::Warn,"Good message");
+    } else if (frame.m_status == UWBRxStatus::ERROR) {
+        logger.log(LogLevel::Warn, "Message in error");
```
xgroleau, line 11: be more verbose in the logging 

"InterlocManager: UWB error message"

same for the other logging under it

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -1,7 +1,26 @@
 #include "InterlocManager.h"
 #include "bsp/BSPContainer.h"
 #include <Task.h>
+#define NB_CALIB_TRIES 10
 
+bool isFrameGood(UWBRxFrame frame, ILogger& logger) {
```
xgroleau, line 6: isFrameOk, also you could make the function as a private method of the interlocmanager, would avoid passing the logger

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -18,27 +37,24 @@ void InterlocManager::startInterloc() {
...
-    UWBRxFrame rxFrame;
-
-    while (true) {
-        m_decaB.receiveAsync(rxFrame, 0);
-        m_decaA.transmit(data, sizeof(data));
-
-        while (rxFrame.m_status == UWBRxStatus::ONGOING) {
-            Task::delay(1);
+    m_decaB.setState(DW_STATE::RESPOND_CALIB);
+    //    m_decaB.setState(DW_STATE::SEND_CALIB);
```
xgroleau, line 41: remove or keep?

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+        Task::delay(100);
+    }
+}
+
+void InterlocManager::startCalibSingleRespond(uint16_t destinationId, Decawave& device) {
+    int32_t error;
+    int32_t dwCountOffset;
+    // distance between devices in calibration mode in centimeters
+
+    device.setChannel(UWBChannel::CHANNEL_2);
```
xgroleau, line 120: should the channel be hardcoded? do we want a compilation setting or at least a define?

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+    device.setTxAntennaDLY(DEFAULT_TX_ANT_DLY);
+    device.setRxAntennaDLY(DEFAULT_RX_ANT_DLY);
+
+    int i = 0;
+    double val = 0;
+    while (device.getState() == DW_STATE::RESPOND_CALIB) { // find exit condition
+        val += receiveTWRSequence(destinationId, device);
+        i++;
+        if (i > NB_CALIB_TRIES - 1) {
```
xgroleau, line 129: So it's the number of measurements not the number of tries/retries is that it?

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+
+            i = 0;
+            // Compute the distance error between acquired and actual
+            error = (int32_t)((val * 100 / NB_CALIB_TRIES) - m_distanceForCalibCm);
+            val = 0;
+            // convert distance error to tick count in DW device
+            // P type control, P  = 0.9
+            // TODO could had a PID type control #futureImprovement
```
xgroleau, line 137: `could add a PID`

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+            dwCountOffset = 0.9 * error * DW_INTERNAL_CLOCK_RFEQ / SPEED_OF_LIGHT / 100;
+            // antenna delay is considered equal on Rx and Tx
+            device.setTxAntennaDLY((dwCountOffset >> 1) + device.getTxAntennaDLY());
+            device.setRxAntennaDLY((dwCountOffset >> 1) + device.getRxAntennaDLY());
+            if (dwCountOffset >> 1 <= 1) {
```
xgroleau, line 142: I'm not sure when this check is true or false

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+    uint64_t respTxTs;
+    uint64_t finalRxTs;
+    UWBMessages::TWRResponse responseMsg{};
+    double varA;
+    double varB;
+    double varC;
+    double varD;
+    double distanceEval;
+    double tof;
+    uint64_t tofDtu;
```
xgroleau, line 163: dont declare then if you don't initiate them. you should declare them on their first value. so at line 175 for pollRxTs for example

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+    // wait for Final message reception
+    device.receive(rxFrame, 0);
+    if (!isFrameGood(rxFrame, m_logger)) {
+        return 0;
+    }
+    if (rxFrame.m_frame->m_functionCode != UWBMessages::FunctionCode::TWR_FINAL) {
+        return 0;
+    }
+
+    UWBMessages::TWRFinal* finalTWRFrame = (UWBMessages::TWRFinal*)(rxFrame.m_frame);
```
xgroleau, line 194: use static_cast<UWBMessages::TWRFinal*>(rxFrame.m_frame);

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+    device.getTxTimestamp(&respTxTs);
+    finalRxTs = rxFrame.m_rxTimestamp;
+
+    // evaluate distance
+    varA = (double)finalTWRFrame->m_respMinPoll;
+    varB = (double)((uint32_t)finalRxTs - (uint32_t)respTxTs);
+    varC = (double)(finalTWRFrame->m_finaleMinResp);
+    varD = (double)((uint32_t)respTxTs - (uint32_t)pollRxTs);
```
xgroleau, line 202: why cast them to uint32_t? couldn't we loose information? If not, why are they uint64_t to begin with?

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+    UWBRxFrame responseFrame;
+    UWBRxFrame finalFrame;
+
+    UWBMessages::TWRPoll pollMsg{};
+    UWBMessages::TWRFinal finalMsg{};
+
+    uint64_t responseTimestamp;
+    uint64_t pollTimestamp;
+    uint64_t finalTxTime;
+    uint64_t finalTxTs;
```
xgroleau, line 223: don't declare uninitialized variable unless you need to (eg. a while loop), initialize them at the same time you write on them for the first time.

src/bsp/include/bsp/IInterlocManager.h
```diff
@@ -15,6 +15,27 @@ class IInterlocManager {
      */
     virtual void startInterloc() = 0;
 
+    /**
+     * @brief Start calibration in initiator mode, it polls in TWR and send the final msg
+     */
+    virtual void startCalibSingleInit() = 0;
```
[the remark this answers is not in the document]
cquesnel: There are two actors in calibration: one initator and one reposponder. The init here stands for initiator. I would suggest changing the name for `startCalibSingleInitiator()` just so it's less confusing

src/bsp/include/bsp/IInterlocManager.h
```diff
@@ -15,6 +15,27 @@ class IInterlocManager {
...
+    /**
+     * @brief Start calibration in responder mode, it sends the response msg and compute the
+     * distance
+     */
+    virtual void startCalibSingleRespond() = 0;
+
+    /**
+     * @brief Sets the targeted distance for calibration
+     */
+    virtual void setCalibDistance(uint16_t distanceForCalibCm) = 0;
```
cquesnel, line 32: The documentation should document the argument

src/bsp/src/stm32/include/Decawave.h
```diff
@@ -125,6 +139,81 @@ class Decawave {
...
+     * @param state The next state of calibration
+     */
+    void setState(DW_STATE state);
+
+    /**
+     * @brief Sets a 5 byte timestamp into the specied message field
+     * @param ts Timestamp to insert in message
+     * @param tsField Field to put timestamp in
+     */
+    void finalMsgAddTs(uint8* tsField, uint64_t ts);
```
cquesnel, line 216: Not sure this belongs in the decawave class. It is specific for a single message.

src/bsp/src/stm32/include/InterlocManager.h
```diff
@@ -24,7 +28,13 @@ class InterlocManager : public IInterlocManager {
     Decawave m_decaB;
 
     uint8_t m_sequenceID = 0;
-
+    uint16_t m_distanceForCalibCm = 75;
```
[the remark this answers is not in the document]
cquesnel: If we use a default calibration distance, move it to a constant somewhere

src/bsp/src/stm32/include/UWBMessages.h
```diff
@@ -34,11 +34,17 @@ namespace UWBMessages {
         FunctionCode m_functionCode;
     };
 
+    struct TWRPoll {
+        DWFrame m_header_frame;
+    };
+
     struct TWRResponse {
+        DWFrame m_header_frame;
         uint64_t m_calculatedTOF;
```
cquesnel, line 43: Is the calculated TOF still used? If not, remove it to save airtime

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -1,4 +1,5 @@
 #include "Decawave.h"
+#include "hal/hal_gpio.h"
```
cquesnel, line 2: Unused header

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -272,4 +289,49 @@ void Decawave::retrieveRxFrame(UWBRxFrame* frame) {
...
+void Decawave::getRxTimestamp(uint64_t* rxTimestamp) {
+    deca_selectDevice(m_spiDevice);
+    dwt_readrxtimestamp((uint8_t*)rxTimestamp);
+}
+void Decawave::getSysTime(uint64_t* sysTime) {
+    deca_selectDevice(m_spiDevice);
+    dwt_readrxtimestamp((uint8_t*)sysTime);
+}
+
+void Decawave::finalMsgAddTs(uint8* tsField, uint64_t ts) {
```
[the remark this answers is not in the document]
cquesnel: Decawave made typedefs for all uintx_t... So anywhere we include their driver, both work

src/bsp/src/stm32/src/Decawave.cpp
```diff
@@ -272,4 +289,49 @@ void Decawave::retrieveRxFrame(UWBRxFrame* frame) {
...
+    dwt_readtxtimestamp((uint8_t*)txTimestamp);
+}
+
+void Decawave::getRxTimestamp(uint64_t* rxTimestamp) {
+    deca_selectDevice(m_spiDevice);
+    dwt_readrxtimestamp((uint8_t*)rxTimestamp);
+}
+void Decawave::getSysTime(uint64_t* sysTime) {
+    deca_selectDevice(m_spiDevice);
+    dwt_readrxtimestamp((uint8_t*)sysTime);
```
cquesnel, line 315: Shouldn't this get the sys time?

src/bsp/src/stm32/src/Decawave.cpp
cquesnel, line 318: This isn't only for the final message, it is a utils method to convert a TS to it's binary value and it should be named as such (and as I said in the .h, not sure this belongs here)

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+    }
+}
+
+double InterlocManager::receiveTWRSequence(uint16_t destinationId, Decawave& device) {
+    UWBRxFrame rxFrame;
+    uint64_t pollRxTs;
+    uint64_t respTxTs;
+    uint64_t finalRxTs;
+    UWBMessages::TWRResponse responseMsg{};
+    double varA;
```
cquesnel, line 157: Do we have better names for these?

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+    if (rxFrame.m_frame->m_functionCode != UWBMessages::FunctionCode::TWR_POLL) {
+        return 0;
+    }
+
+    pollRxTs = rxFrame.m_rxTimestamp;
+
+    // construct and send Response message
+    constructUWBHeader(destinationId, UWBMessages::DATA, UWBMessages::TWR_RESPONSE,
+                       (uint8_t*)&responseMsg, sizeof(responseMsg));
+    //    respTxTs = (pollRxTs + (POLL_RX_TO_RESP_TX_DLY_UUS * UUS_TO_DWT_TIME))>>8;
```
cquesnel, line 180: If you couldn't get the delayed working, add a todo here so we can try again later as this will be needed

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+    double varD;
+    double distanceEval;
+    double tof;
+    uint64_t tofDtu;
+
+    // receive poll message
+    device.receive(rxFrame, 0);
+
+    if (!isFrameGood(rxFrame, m_logger)) {
+        return 0;
```
cquesnel, line 169: return -1 if there is an error. That way, we can catch it at a higher level

src/bsp/src/stm32/src/InterlocManager.cpp
```diff
@@ -66,3 +82,174 @@ bool InterlocManager::constructUWBHeader(uint16_t destinationId,
...
+        return false;
+    }
+
+    //  Retrieve poll transmission and response reception timestamp.
+    device.getTxTimestamp(&pollTimestamp);
+    responseTimestamp = responseFrame.m_rxTimestamp;
+
+    //  Compute final message transmission time
+    finalTxTime = (responseTimestamp + (RESP_RX_TO_FINAL_TX_DLY_UUS * UUS_TO_DWT_TIME)) >> 8;
+    finalTxTs = (finalTxTime << 8); //+device.getTxAntennaDLY();
```
cquesnel, line 246: Remove the commented section if it is not to be used

cquesnel (changes requested):

src/bsp/src/stm32/src/InterlocManager.cpp
dubh3401: I removed the cast,
yes we loose information, but that's the point, it prevents breaking the system if the  clock wraps mid communication

src/bsp/src/stm32/src/InterlocManager.cpp
dubh3401: created DEFAULT_CHANNEL is Decawave.h

src/bsp/src/stm32/src/InterlocManager.cpp
dubh3401: I prefer to let it there as a reminder that the delays are not properly set. More reading needs to be done in the following document : https://www.google.com/search?channel=fs&client=ubuntu&q=DecaRanging+ARM+Source+Code+Guide

src/bsp/src/stm32/include/UWBMessages.h
dubh3401: let for futur use, see https://www.google.com/search?channel=fs&client=ubuntu&q=DecaRanging+ARM+Source+Code+Guide

src/bsp/src/stm32/include/UWBMessages.h
```diff
@@ -35,16 +35,16 @@ namespace UWBMessages {
...
 
     struct TWRResponse {
-        DWFrame m_header_frame;
+        DWFrame m_headerFrame;
         uint64_t m_calculatedTOF;
     };
 
-    struct TWRFinal {
-        DWFrame m_header_frame;
+    struct TWRFinal : DWFrame {
```
cquesnel, line 46: not sure you wan't to inherit here. This will double the size of the message...

cquesnel (approved):

src/bsp/src/stm32/include/UWBMessages.h
dubh3401: oops no not supposed to be there

xgroleau (approved):

PR #64, opened 2021-04-01
Feat/swarinfo 308 calib api

cquesnel:
Added Interloc Message Handler that can start/stop a calibration


src/Main.cpp
```diff
@@ -190,24 +192,48 @@ class CommMonitoringTask : public AbstractTask<4 * configMINIMAL_STACK_SIZE> {
...
+        AbstractTask(taskName, priority) {}
 
-    ~InterlocTask() override = default;
+    ~HardwareInterlocTask() override = default;
 
   private:
     void task() override { BSPContainer::getInterlocManager().startInterloc(); }
 };
 
+class SoftwareInterlocTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
```
xgroleau, line 206: more of a "InterlocMessageHandlerTask " or there will be more processing here eventually?

src/interloc/src/InterlocMessageHandler.cpp
```diff
@@ -0,0 +1,101 @@
...
+        return false;
+    }
+
+    if (std::holds_alternative<StopCalibrationDTO>(messageVariant)) {
+        m_interlocManager.stopCalibration();
+        return true;
+    }
+
+    if (const auto* setDistanceMsg = std::get_if<SetCalibrationDistanceDTO>(&messageVariant)) {
+        m_interlocManager.setCalibDistance(setDistanceMsg->getDistance() * 100);
```
xgroleau, line 74: *100 for cm -> m? if so is there a reason why one is in cm and one in m?

xgroleau (approved):
Honestly GJ, I don't have much to say! 👍

src/Main.cpp
cquesnel: I was thinking of putting a process() for the Interloc class which could handle filtering and such

src/interloc/src/InterlocMessageHandler.cpp
cquesnel: More m->cm, but honestly the only reason is I found out the function took cm instead of meters and we are using meters everywhere else in the code, so I left it that way

PR #74, opened 2021-04-09
Feat/swarinfo 353 expose interloc api to host

xgroleau:
Can now manage interloc api with the host

### DIFF
--- a/src/message-handler/src/HiveMindHostApiRequestHandler.cpp
+++ b/src/message-handler/src/HiveMindHostApiRequestHandler.cpp
@@ -29,11 +29,17 @@ bool HiveMindHostApiRequestHandler::handleHiveMindHostApiRequest(
 
     if (const auto* neighborReq = std::get_if<GetNeighborRequestDTO>(&request.getRequest())) {
         uint16_t id = neighborReq->getNeighborId();
-        std::optional<RelativePosition> pos = m_interloc.getRobotPosition(id);
-        NeighborPositionDTO posDTO(pos->m_distance, pos->m_relativeOrientation,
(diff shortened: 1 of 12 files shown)


t-constantin (approved):
LGTM!

src/message-handler/src/HiveMindHostApiRequestHandler.cpp
```diff
@@ -20,11 +21,48 @@ bool HiveMindHostApiRequestHandler::handleRequest(const MessageDTO& message) {
...
+
+    if (std::holds_alternative<GetNeighborsListRequestDTO>(request.getRequest())) {
+
+        GetNeighborsListResponseDTO neighborResp(NULL, 0);
+        const PositionsTable& posTable = m_interloc.getPositionsTable();
+
+        neighborResp.setRawNeighborsLength(posTable.m_positionsLength);
+        auto& rawNeighbors = neighborResp.getRawNeighbors();
+        for (uint16_t i = 0; i < neighborResp.getNeighborsLength(); i++) {
+            rawNeighbors[i] = posTable.m_positions[i].m_robotId;
```
cquesnel, line 57: If a robot is not in LOS but is still present in the table because it is still in range of the DW, is it considered a neighbour?

cquesnel (approved):

src/message-handler/src/HiveMindHostApiRequestHandler.cpp
xgroleau: As of now yes it does. If we found the need to only have the ones in LOS, we could add a field in the request  to filter and send only those in LOS.

PR #75, opened 2021-04-09
Feat/swarinfo 352 use hashmap for faster lookup

xgroleau:


src/bittybuzz/src/BittyBuzzClosureRegister.cpp
```diff
@@ -3,18 +3,21 @@
...
+    m_closureHeapIdx(closureHeapIdx),
+    m_selfHeapIdx(selfHeapIdx),
+    m_description(std::move(description)) {}
 
 bool BittyBuzzClosureRegister::registerClosure(const char* functionName,
                                                bbzheap_idx_t closureHeapIdx,
                                                bbzheap_idx_t selfHeapIdx,
                                                const BittyBuzzFunctionDescription& description) {
-    if (m_closureRegistersLength >= m_maxSize) {
+    if (m_closureNameRegisters.isFull() || m_closureRegisterMap.isFull()) {
```
t-constantin, line 20: To we want to keep older values or insert newer ones?

src/bittybuzz/src/BittyBuzzClosureRegister.cpp
```diff
@@ -29,49 +32,38 @@ bool BittyBuzzClosureRegister::registerClosure(const char* functionName,
...
-
-    for (uint16_t i = 0; i < m_closureRegistersLength; i++) {
-        size_t hash = std::get<0>(m_closureRegisters[i]);
-        if (functionNameHash == hash) {
-            return std::get<1>(m_closureRegisters[i]);
-        }
-    }
-
-    return {};
+    return m_closureRegisterMap.at(functionName);
```
t-constantin, line 55: usually a hashmap throws when using at with somehting not in the map, is this the case for a cpp-common/HashMapStack ?

src/bittybuzz/src/BittyBuzzClosureRegister.cpp
```diff
@@ -29,49 +32,38 @@ bool BittyBuzzClosureRegister::registerClosure(const char* functionName,
...
 }
 
 std::optional<std::reference_wrapper<const BittyBuzzRegisteredClosure>> BittyBuzzClosureRegister::
     getRegisteredClosure(uint16_t idx) const {
-    if (idx > m_closureRegistersLength - 1) {
-        return {};
+    auto functionNameOpt = m_closureNameRegisters.at(idx);
```
t-constantin, line 60: idem as last comment

src/bittybuzz/src/BittyBuzzClosureRegister.cpp
```diff
@@ -29,49 +32,38 @@ bool BittyBuzzClosureRegister::registerClosure(const char* functionName,
...
+    if (functionNameOpt) {
+        return m_closureRegisterMap.at(functionNameOpt.value().get());
```
t-constantin, line 62: idem

t-constantin (approved):
Few questions but more of my understanding than the actual code.

src/bittybuzz/src/BittyBuzzClosureRegister.cpp
xgroleau: Nope, it returns an optional on our implementation

src/bittybuzz/src/BittyBuzzClosureRegister.cpp
xgroleau: Removed so we can overwrite existing even if the hashmap is full

PR #76, opened 2021-04-11
Feat/swarinfo 254 simulation interloc

cquesnel:


src/bsp/src/posix/src/InterlocManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+    const geometry_msgs::TransformStamped& currentAgentHBTransform,
+    const geometry_msgs::Pose& distantAgentPose,
+    const geometry_msgs::TransformStamped& distantAgentHBTransform) {
+
+    // Convert the poses into the HB frame
+    geometry_msgs::Pose currentAgentPoseHBFrame;
+    geometry_msgs::Pose distantAgentPoseHBFrame;
+
+    tf2::doTransform(currentAgentPose, currentAgentPoseHBFrame, currentAgentHBTransform);
+    tf2::doTransform(distantAgentPose, distantAgentPoseHBFrame, distantAgentHBTransform);
```
etienn8, line 30: If its not so heavy, you could add in which frame each pose is in. This helps the reader know if tranformation can be done. For example the AgentHBTransform let transform a Pose from the robot Frame into the Hiveboard frame. So adding the suffix RobotFrame after the currentAgentPose so developper have an easier time understanding the process of the function. This could be add a comment in the .h though

etienn8: Actually from the ModelState Message we have the pose the in world frame. Maybe I'm missing something but should be the currentAgentPose inverted?

etienn8 (commented):
Nice work! I dont

etienn8 (approved):

src/bsp/src/posix/include/InterlocManager.h
```diff
@@ -1,33 +1,59 @@
...
 
     void setPositionUpdateCallback(positionUpdateCallbackFunction_t callback,
-                                   void* context) override {
-        // TODO: implement
-        (void)callback;
-        (void)context;
-    };
+                                   void* context) override;
 
+    // Calib API is not needed in simulation. Can just ignore any calls to these functions
```
xgroleau, line 21: Even if there is no logic here, you could put the definition of the function in the .cpp so the compilation time is reduced.

src/bsp/src/posix/src/InterlocManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+    try {
+        transform = m_tfBuffer.lookupTransform(agentName + HIVEBOARD_SUFFIX,
+                                               agentName + ROBOT_CENTER_SUFFIX, ros::Time(0));
+    } catch (tf2::TransformException& ex) {
+        m_logger.log(LogLevel::Warn, "Could not find transform to hiveboard for %s",
+                     agentName.c_str());
+        m_logger.log(LogLevel::Warn, "%s", ex.what());
+    }
+
+    return transform;
```
xgroleau, line 68: So garbage is returned if there is an exception? Why not return {} since it's an optinal?

src/bsp/src/posix/src/InterlocManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+            // Do not calculate interloc with ourself
+            continue;
+        }
+
+        tf2::Transform relTransform = computeRelativeTransform(
+            msg.pose[idxLookup[currentAgentId]], m_baseLinkToHiveBoardTransforms[currentAgentId],
+            msg.pose[index], m_baseLinkToHiveBoardTransforms[agentId]);
+
+        double distance = getDistance(relTransform);
+        double orientation = getOrientation(relTransform) * 180 / M_PI;
```
xgroleau, line 115: I was wondering, is there a reason why we are in degrees and not rad? Most sine and cosine functions uses rad. Is it since it's easier for precision? (i.e. more precise with fixed point)

xgroleau (approved):
GJ! Juste couple questions

src/bsp/src/posix/src/InterlocManager.cpp
cquesnel: My bad, I forgot to return inside of the catch

src/bsp/src/posix/src/InterlocManager.cpp
cquesnel: I was using degrees just because it is easier for us to understand, but we could switch everything to rads if you want to

cquesnel (commented):
x

PR #81, opened 2021-04-14
Feat/swarinfo 367 bbvm math functions

xgroleau:
We can now register "libraries" in buzz, wither global (so registered on the global scope) or via table id (the  lib will be in a table, used as a namespace) 

- Added a way to add table libraries to buzz (only one one level libraries for now)
- Migrate the global function to the library registration on the global scope
- Added a math library for basic functionalities
- The vm init process is separated from the construction, eventually we may want to be able to reset the vm


src/bittybuzz/src/BittyBuzzSystem.cpp
```diff
@@ -5,6 +5,7 @@ IUserInterface* BittyBuzzSystem::g_ui = NULL;
 const IBittyBuzzStringResolver* BittyBuzzSystem::g_stringResolver = NULL;
 IBittyBuzzClosureRegister* BittyBuzzSystem::g_closureRegister = NULL;
 IBittyBuzzMessageService* BittyBuzzSystem::g_messageService = NULL;
+IBSP* BittyBuzzSystem::g_bsp = NULL;
```
g-lesssard, line 8: This could be set with the container instead of setting it to null

src/bittybuzz/include/bittybuzz/BittyBuzzSystem.h
```diff
@@ -41,6 +42,11 @@ namespace BittyBuzzSystem {
      **/
     extern IBittyBuzzMessageService* g_messageService;
 
+    /**
+     *@brief BSP used by the bbvm for random numbers
+     **/
+    extern IBSP* g_bsp;
```
g-lesssard, line 48: Why the global variable? Could you change it to a static reference to the value obtained from the BSP:Container?

src/bittybuzz/src/BittyBuzzMathFunctions.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+    /* Compare them and return the bigger one */
+    int cmp = bbztype_cmp(a, b);
+    bbzvm_lload(cmp >= 0 ? 1 : 2);
+    bbzvm_ret1();
+}
+
+void BittyBuzzMathFunctions::bbzmath_rng_uniform() {
+    bbzvm_assert_lnum(0); // NOLINT
+    uint32_t randomNum = BittyBuzzSystem::g_bsp->generateRandomNumber();
```
g-lesssard, line 309: Could it be changed to get the reference from the container? would be safer than linking to an external pointer with no null check...

g-lesssard (changes requested):

g-lesssard (changes requested):
Not a huge fan of the global variable to the bsp... Why not use the container instead? Also, not sure about the template. Would it be simpler to create an interface for the type of the iterator that you need for the for each?

src/bittybuzz/include/bittybuzz/BittyBuzzLib.h
```diff
@@ -0,0 +1,33 @@
+#ifndef __BITTYBUZZSTDLIB_H_
+#define __BITTYBUZZSTDLIB_H_
+
+#include "IBittyBuzzLib.h"
+#include <cstdint>
+#include <optional>
+
+template <typename Container>
```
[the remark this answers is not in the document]
xgroleau: There is an interface, IBittyBuzzlLib. the template is only for the container of BittyBuzzLibMemberRegister so we can store it somewhere. I could have done a static array and then copy it, though this array would have a max size. By using template, the object will be exactly the size it needs to be, not more, not less

src/bittybuzz/src/BittyBuzzSystem.cpp
xgroleau: We should never use the container outside the main. Always inject the dependencies. If we use the container, we can't mock the BSP. The reason I use global variables here are because bittybuzz has no context, so the only way I can keep context is via global variables

src/bittybuzz/include/bittybuzz/BittyBuzzSystem.h
xgroleau: See the other comment.

src/bittybuzz/src/BittyBuzzMathFunctions.cpp
xgroleau: No it can't since the namespace cannot initialize a reference, it has no constructor. We need to inject it via the bbvm so it's possible to mock. Also this pointer should never be null since this is only called after the bbvm is initialized, and the bbvm takes a reference at it's constructor to initialize this pointer. In other words. the BittyBuzzSystem is initialized by the BittyBuzzVm.

src/bittybuzz/include/bittybuzz/BittyBuzzLib.h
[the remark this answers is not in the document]
g-lesssard: From what I understand, you do a for each on the elements of the container. It's just a bit confusing since as you said, I could be trying to set a static array in the template, but since you are using a for each in your code it would not work. Having an interface would make it more explicit to the user that the Container is expected to have .begin() and .end() of type BittyBuzzLibMemberRegister.

It's more of a nitpick, the template makes the usage a bit complicated and lacks transparency.

src/bittybuzz/src/BittyBuzzSystem.cpp
g-lesssard: Could we mock the container? I just find it dangerous to have global pointers that are initialized as null... Also, the notation is a bit confusing, you write your variables BittyBuzzSystem::g_xxxx but since the variable is in the namespace, it is not global. 

Although, If we can't mock the container, then I would keep it as is since I prefer having unit tests on this compared to avoiding global pointers

g-lesssard (approved):

src/bittybuzz/src/BittyBuzzMathFunctions.cpp
```diff
@@ -0,0 +1,311 @@
...
+    // Get args
+    float arg = 0;
+    bbzobj_t* o = bbzheap_obj_at(bbzvm_locals_at(1)); // NOLINT
+    if (bbztype_isfloat(*o)) {
+        arg = bbzfloat_tofloat(o->f.value);
+    } else if (bbztype_isint(*o)) {
+        arg = o->i.value;
+    } else {
+        bbzvm_seterror(BBZVM_ERROR_MATH);
+        return;
```
etienn8, line 34: It seems that you repeat this code section in each math function expect the abs one. Is there a reason why you didn't create a function getArg() instead?

etienn8 (commented):
Nice job! Small comment on the math section

src/bittybuzz/src/BittyBuzzMathFunctions.cpp
xgroleau: Done

PR #82, opened 2021-04-14
Feat/swarinfo 364 state tracer

cquesnel:
Added a way to debug state transitions by logging the transitions as well a a transition event in a queue.

### DIFF
--- a/src/bsp/src/stm32/src/interloc/src/InterlocManager.cpp
+++ b/src/bsp/src/stm32/src/interloc/src/InterlocManager.cpp
@@ -41,8 +41,8 @@ void InterlocManager::startInterloc() {
     syncClocks();
 
     // Uncomment one of the following lines to go into TWR
-    // m_stateHandler.setState(InterlocStates::WAIT_POLL);
-    // m_stateHandler.setState(InterlocStates::SEND_POLL);
+    // m_stateHandler.setState(InterlocStates::WAIT_POLL, InterlocEvent::NO_EVENT);
+    m_stateHandler.setState(InterlocStates::SEND_POLL, InterlocEvent::NO_EVENT);
 
     while (true) {
         m_stateHandler.process();
--- a/src/bsp/src/stm32/src/interloc/src/InterlocManager.cpp
+++ b/src/bsp/src/stm32/src/interloc/src/InterlocManager.cpp
@@ -42,7 +42,7 @@ void InterlocManager::startInterloc() {
 
     // Uncomment one of the following lines to go into TWR
     // m_stateHandler.setState(InterlocStates::WAIT_POLL, InterlocEvent::NO_EVENT);
-    m_stateHandler.setState(InterlocStates::SEND_POLL, InterlocEvent::NO_EVENT);
+    // m_stateHandler.setState(InterlocStates::SEND_POLL, InterlocEvent::NO_EVENT);
 
     while (true) {
         m_stateHandler.process();
--- a/src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
+++ b/src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
@@ -11,11 +11,8 @@ void InterlocStateHandler::setState(InterlocStates state, InterlocEvent event) {
         m_stateTracer.pop();
     }
 
-    auto& stateTransition = m_stateTracer.getNextAllocation().value().get();
-    stateTransition.m_fromState = m_currentStateName;
-    stateTransition.m_toState = state;
-    stateTransition.m_transitionEvent = event;
-    m_stateTracer.advance();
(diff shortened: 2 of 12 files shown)


src/bsp/src/stm32/src/interloc/src/InterlocManager.cpp
```diff
@@ -41,8 +41,8 @@ void InterlocManager::startInterloc() {
     syncClocks();
 
     // Uncomment one of the following lines to go into TWR
-    // m_stateHandler.setState(InterlocStates::WAIT_POLL);
-    // m_stateHandler.setState(InterlocStates::SEND_POLL);
+    // m_stateHandler.setState(InterlocStates::WAIT_POLL, InterlocEvent::NO_EVENT);
+    // m_stateHandler.setState(InterlocStates::SEND_POLL, InterlocEvent::NO_EVENT);
```
stlaperriere, line 45: Commented code?

src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
```diff
@@ -2,9 +2,22 @@
...
-    m_currentState(&InterlocStateContainer::getState(InterlocStates::DEFAULT)) {}
+    m_currentStateName(InterlocStates::DEFAULT),
+    m_currentState(&InterlocStateContainer::getState(InterlocStates::DEFAULT)),
+    m_stateTracer(m_stateTracerData.data(), MAX_TRACER_TRANSITIONS) {}
 
-void InterlocStateHandler::setState(InterlocStates state) {
+void InterlocStateHandler::setState(InterlocStates state, InterlocEvent event) {
+    if (m_stateTracer.isFull()) {
+        m_stateTracer.pop();
+    }
```
stlaperriere, line 12: Why the need to pop if m_stateTracer is a circular queue?

src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
```diff
@@ -2,9 +2,22 @@
...
+
+    auto& stateTransition = m_stateTracer.getNextAllocation().value().get();
+    stateTransition.m_fromState = m_currentStateName;
+    stateTransition.m_toState = state;
+    stateTransition.m_transitionEvent = event;
+    m_stateTracer.push(stateTransition);
```
stlaperriere, line 18: You push the same value you just got from the queue? Won't this just invariably overwrite the last value?

stlaperriere (changes requested):
I have questions.

src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
cquesnel: The queue won't push if it's full. So we clear the oldest data before pushing

src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
cquesnel: The getNextAllocation() returns a reference to the next empty slot in the queue, so I change the data in that slot. The push then just changes the counters in the queue. The push uses a reference and doesn't copy the data, so if I didn't use the getNextAllocation() first, the data I push would go out of scope the moment I exit the function.

src/bsp/src/stm32/src/interloc/src/InterlocManager.cpp
cquesnel: Yes, this starts the interloc in one of two modes, but won't work if no hivesight is on the board so it is just there to make it easier when testing without having to always rewrite these lines

src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
stlaperriere: Oh okay I see.

stlaperriere (approved):

src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
xgroleau: The push is supposed to copy the data though, as of now you are copying twice and not calling the destructor when overwriting it (which is not an issue in this case)

xgroleau (changes requested):

src/bsp/src/stm32/src/interloc/src/InterlocStateHandler.cpp
xgroleau: getNextAllocation should only be used with advance afterwards. push == getNextAllocation + advance. It's just if you want the  variable temporarly and the confirm if you advance (or not), you can use getNextAllocation

xgroleau (approved):

PR #83, opened 2021-04-15
Feat/swarinfo 358 twr fsm

dubh3401:


src/bsp/src/stm32/src/interloc/src/TwoWayRanging.cpp
```diff
@@ -36,7 +36,7 @@ std::optional<double> TwoWayRanging::calculateDistance() const {
     double tof = tofDtu * DWT_TIME_UNITS;
     double distance = tof * SPEED_OF_LIGHT;
 
-    if (distance > 0 && distance < MAX_POSSIBLE_DISTANCE) {
+    if (distance < MAX_POSSIBLE_DISTANCE) {
```
cquesnel, line 39: Revert back to the double check because sometimes, the calculation fails and 0 is returned as a distance which isn't good

cquesnel (approved):

PR #84, opened 2021-04-15
Feat/swarinfo 386 notification queue

xgroleau:
Uses notification queu for waiting messages

### DIFF
--- a/src/Main.cpp
+++ b/src/Main.cpp
@@ -55,7 +55,7 @@ class BittyBuzzTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
         auto bbzFunctions = BittyBuzzFactory::createBittyBuzzGlobalLib();
         auto mathLib = BittyBuzzFactory::createBittyBuzzMathLib();
         std::array<std::reference_wrapper<IBittyBuzzLib>, 2> buzzLibraries{{bbzFunctions, mathLib}};
-        if (m_bittybuzzVm.init(buzzLibraries.data(), buzzLibraries.size())) {
+        if (!m_bittybuzzVm.init(buzzLibraries.data(), buzzLibraries.size())) {
             m_logger.log(LogLevel::Error, "BBZVM failed to initialize. state: %d err: %d",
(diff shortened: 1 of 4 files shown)


src/Main.cpp
```diff
@@ -225,10 +231,15 @@ class SoftwareInterlocTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE>
...
     // TODO: remove once it is used somewhere else
     IInterloc& m_interloc;
+    NotificationQueue<MessageDTO>& m_interlocMessageQueue;
     IInterlocMessageHandler& m_interlocMessageHandler;
 
     void task() override {
         while (true) {
+            // Verify that we have a message to process
+            if (m_interlocMessageQueue.isEmpty()) {
+                m_interlocMessageQueue.wait(UINT32_MAX);
```
g-lesssard, line 241: Do we really want to be blocking? We could make the delay relatively long, but still have a timeout in case the notifications fails for some reason. If we keep it blocking, we should use portMAX_DELAY. For posix, I think we have a define for it in propolis, and for freeRTOS it would use the one for the port and would be safer

g-lesssard (approved):
LGTM, minor comment about the wait time

src/Main.cpp
xgroleau: I'm not sure why we wouldn't want to block, we added that to  avoid useless processing. This removes the need for it. The notification shouldn't fail, if they do, we have bigger underlying problems with the design/code. portMAX_DELAY is the same thing as UINT32_MAX since all our builds are at least 32 bits.

cquesnel: I agree with @g-lesssard on this. Checking if the queue has something in it every few ms won't impact performance at all while possibly preventing bugs with edge cases we haven't thought of.

cquesnel (approved):

src/Main.cpp
xgroleau: Changed it to 500 ms. though generally "partially hiding" bugs aren't the way to go  since it becomes hard to reproduce bugs and can hide bigger inner problems in the code.

PR #85, opened 2021-04-16
Tx bloquant

dubh3401:


src/bsp/src/stm32/src/interloc/src/Decawave.cpp
```diff
@@ -17,6 +17,19 @@ void Decawave::rxCallback(const dwt_cb_data_t* callbackData, void* context) {
...
+    BaseType_t taskWoken = pdFALSE;
+
+    (void)callbackData;
+
+    if (static_cast<Decawave*>(context)->m_rxTaskHandle != NULL) {
+        vTaskNotifyGiveFromISR(static_cast<Decawave*>(context)->m_rxTaskHandle, &taskWoken);
+    }
+
+    static_cast<Decawave*>(context)->m_rxTaskHandle = NULL;
+    portYIELD_FROM_ISR(taskWoken);
```
stlaperriere, line 30: You could simplify this code by declaring a pointer only once on m_rxTaskHandle instead of re-writing `static_cast<Decawave*>(context)->m_rxTaskHandle` three times.

stlaperriere (approved):
Just a little nitpick.

src/bsp/src/stm32/src/interloc/src/Decawave.cpp
dubh3401: Good suggestion, but adding another var takes more memory and time than a simple cast

PR #87, opened 2021-04-18
Feat/swarinfo 250 spi driver finishing touches

g-lesssard:


src/bsp/src/stm32/include/SpiHeader.h
```diff
@@ -24,10 +24,11 @@ namespace EspHeader {
         uint8_t rawValue;
     };
 
-    struct Header {
+    struct __attribute__((__packed__)) Header {
```
xgroleau, line 27: Why do we need the header packed here?

src/bsp/src/stm32/src/SpiEsp.cpp
```diff
@@ -25,47 +19,68 @@ SpiEsp::SpiEsp(ICRC& crc, ILogger& logger) :
...
+    // Appending CRC32
+    *(uint32_t*)&m_outboundMessage.m_data[length] =
+        m_crc.calculateCRC32(m_outboundMessage.m_data.data(), length);
+    m_outboundMessage.m_sizeBytes = (uint16_t)(length + CRC32_SIZE);
+    m_txState = transmitState::SENDING_HEADER;
+
+    // Wait for transmission to be over
+    m_sendingTaskHandle = xTaskGetCurrentTaskHandle();
+    // Wait for transmission to be over. Will be notified when ACK received or upon error
+    m_sendingTaskHandle = xTaskGetCurrentTaskHandle();
```
xgroleau, line 58: Line duplication with 56

src/bsp/src/stm32/src/SpiEsp.cpp
```diff
@@ -86,14 +104,20 @@ void SpiEsp::execute() {
         m_inboundHeader = (EspHeader::Header*)m_inboundMessage.m_data.data();
         if (m_inboundHeader->crc8 !=
             m_crc.calculateCRC8(m_inboundHeader, EspHeader::sizeBytes - 1)) {
-            m_logger.log(LogLevel::Error, "Received corrupted SPI ESP header");
+            m_logger.log(LogLevel::Debug, "Received corrupted SPI ESP header");
```
xgroleau, line 107: Is this frequent so you setted it in debug?

src/bsp/src/stm32/src/SpiEsp.cpp
```diff
@@ -102,35 +126,60 @@ void SpiEsp::execute() {
...
         m_rxState = receiveState::RECEIVING_HEADER;
         rxLengthBytes = EspHeader::sizeBytes;
         break;
     case receiveState::ERROR:
         rxLengthBytes = EspHeader::sizeBytes;
         m_rxState = receiveState::RECEIVING_HEADER;
+        CircularBuff_clear(&m_circularBuf);
+        if (m_receivingTaskHandle != nullptr) {
+            xTaskNotifyGive(m_receivingTaskHandle);
+        }
```
xgroleau, line 182: Should we log here?

xgroleau (approved):

src/bsp/src/stm32/include/SpiHeader.h
g-lesssard: To make sure  the compiler does not pad the struct. During development, I needed to add it since the compiler was padding it up to 10 bytes for some reason

src/bsp/src/stm32/src/SpiEsp.cpp
g-lesssard: I added it so if you dev without a HiveSight, or without the ESP flashed with the right firmware, it will not pollute the logs.

PR #91, opened 2021-04-20
 Feat/swarinfo 368 add broadcast support

xgroleau:
Add broadcast support and fixed some bugs. Waiting on [this PR](https://github.com/MISTLab/BittyBuzz/pull/11) to fully test. Note that the buzz code is not valid, was only for testing purpose in simulation, the agnle was the relative angle, not the azimuth.


cmake/bittybuzz/common.cmake
```diff
@@ -6,8 +6,8 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} PARENT_SCOPE)
 function(bittybuzz_fetch)
     FetchContent_Declare(
         bittybuzz
-        GIT_REPOSITORY https://github.com/MISTLab/BittyBuzz
-        GIT_TAG        6ec05e8
+        GIT_REPOSITORY https://github.com/xgroleau/BittyBuzz
+        GIT_TAG        42cd16de3018639b8eb11179c318be8c5938864e
```
xgroleau, line 10: Change to mist lab when PR is merged

src/bittybuzz/src/BittyBuzzMathFunctions.cpp
```diff
@@ -147,7 +147,7 @@ void BittyBuzzMathFunctions::bbzmath_sin() {
         return;
     }
 
-    bbzvm_pushf(bbzfloat_tofloat(Math::sin(arg.value())));
+    bbzvm_pushf(bbzfloat_fromfloat(Math::sin(arg.value() * Math::pi / 180)));
```
cquesnel, line 150: So the buzz scripts will work in degrees? Why is the conversion only done once? Shouldn't the result be converted back before pushing back to the VM?

src/message-handler/src/MessageDispatcher.cpp
```diff
@@ -34,12 +34,19 @@ bool MessageDispatcher::deserializeAndDispatch() {
         }
```
cquesnel, line 34: No tests for these changes?

cquesnel (changes requested):

src/bittybuzz/src/BittyBuzzMathFunctions.cpp
xgroleau: Since our neighbors values are in degrees, I though it would be simpler to have all the functions take degrees instead of rad. Unless we want to migrate all to rad, though we may lose a bit of precision since rads are smaller and idk how float_16 are precise. You are right though, the return value needs to be converted in degrees again.

src/message-handler/src/MessageDispatcher.cpp
xgroleau: fixed

src/bittybuzz/src/BittyBuzzMathFunctions.cpp
cquesnel: Yeah, you're right, I don't know what I was thinking

cquesnel (approved):

PR #93, opened 2021-04-20
Feat/swarinfo 383 log type of error in bbvm

xgroleau:
Added some more logging, the id of the HM is now shown on every print on the ROS build, makes the logging more relevant on the simulation. Buzz now logs it's error as a string. The log level is now printed, allowing the search or filtering easier using `grep` or `ripgrep`

### DIFF
--- a/src/bittybuzz/include/bittybuzz/BittyBuzzVm.h
+++ b/src/bittybuzz/include/bittybuzz/BittyBuzzVm.h
@@ -43,11 +43,11 @@ class BittyBuzzVm : public IBittyBuzzVm {
 
     bool step() override;
 
-    const char* getState() const override;
+    bbzvm_state getState() const override;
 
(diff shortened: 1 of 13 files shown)


src/bittybuzz/include/bittybuzz/BittyBuzzVm.h
```diff
@@ -43,9 +43,11 @@ class BittyBuzzVm : public IBittyBuzzVm {
 
     bool step() override;
 
-    bbzvm_state getSate() const override;
+    const char* getState() const override;
```
g-lesssard, line 46: Shouldn't getState return the bbzm_state? If not, what's the difference between getState and getStateString?

xgroleau: Reverted to the enum instead

PR #97, opened 2021-04-21
Feat/swarinfo 413 push interloc to buzz

cquesnel:
Make interloc data go up the stack into buzz

### DIFF
--- a/src/bsp/src/stm32/src/interloc/include/interloc/InterlocStateHandler.h
+++ b/src/bsp/src/stm32/src/interloc/include/interloc/InterlocStateHandler.h
@@ -54,8 +54,6 @@ class InterlocStateHandler {
     CircularQueue<StateTransition> m_stateTracer;
 
     uint8_t m_sequenceID = 0;
-
-    // TODO : set manually /AprilTag
     uint16_t m_slotId;
 
     uint8_t m_numFrames = MAX_INTERLOC_SUBFRAMES;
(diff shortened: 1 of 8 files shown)


src/bsp/src/stm32/src/interloc/include/interloc/InterlocStateHandler.h
```diff
@@ -51,9 +54,7 @@ class InterlocStateHandler {
     CircularQueue<StateTransition> m_stateTracer;
 
     uint8_t m_sequenceID = 0;
-
-    // TODO : set manually /AprilTag
```
t-constantin, line 55: How is it handled right now? (Not an issue, just curious)

src/bsp/src/stm32/src/interloc/src/states/src/SetDistanceState.cpp
```diff
@@ -12,8 +12,11 @@ void SetDistanceState::process(InterlocStateHandler& context) {
     if (distance) {
         m_logger.log(LogLevel::Warn, "distance from : %d = %3.3f", context.getCurrentFrameId(),
                      distance.value());
+
+        InterlocBSPContainer::getInterlocManager().updateDistance(
+            context.getBoardIdFromSlotId(context.getCurrentFrameId()), distance.value());
+
         context.setState(InterlocStates::IDLE, InterlocEvent::GOOD_DISTANCE);
-        //        TODO notify Buzz that a new distance value is available
```
t-constantin, line 16: Before removing the to do, is it already done?

t-constantin (approved):
LGTM! Clean code as we like it
Just few questions, but nothing blocking

g-lesssard (approved):
LGTM :+1:

xgroleau (approved):
LGTM

src/bsp/src/stm32/src/interloc/include/interloc/InterlocStateHandler.h
cquesnel: We use the board id (the same that is returned when you greet with the HM). In the next days, we will embed this ID directly in the Nucleo's memory so they will each have a unique ID from 1 to 5

PR #98, opened 2021-04-23
Feat/swarinfo 418 spi stability improvments

g-lesssard:
- Fixes CommMonitoringTask that was exiting on disconnections
- Templated the serializer to choose between accumulator or standard
- Reduced RAM usage of SPI buffers

Less ram used, more stable behavior

### DIFF
--- a/src/Main.cpp
+++ b/src/Main.cpp
@@ -147,13 +147,15 @@ class MessageSenderTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
             HiveMindHostSerializer serializer(*m_stream);
             MessageSender messageSender(m_streamQueue, serializer, BSPContainer::getBSP(),
                                         m_logger);
-            while (m_stream->isConnected()) {
-                // Verify that we have a message to process
-                if (m_streamQueue.isEmpty()) {
-                    m_streamQueue.wait(500);
-                }
-                if (!messageSender.processAndSerialize()) {
-                    m_logger.log(LogLevel::Warn, "Fail to process/serialize in %s", m_taskName);
+            while (true) {
+                if (m_stream->isConnected()) {
+                    // Verify that we have a message to process
+                    if (m_streamQueue.isEmpty()) {
+                        m_streamQueue.wait(500);
+                    }
+                    if (!messageSender.processAndSerialize()) {
+                        m_logger.log(LogLevel::Warn, "Fail to process/serialize in %s", m_taskName);
+                    }
                 }
             }
         }
--- a/src/Main.cpp
+++ b/src/Main.cpp
@@ -120,7 +120,7 @@ class MessageDispatcherTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE>
     }
 };
 
-template <typename SerializerType>
+template <typename SerializerType = HiveMindHostSerializer>
 class MessageSenderTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
   public:
     MessageSenderTask(const char* taskName,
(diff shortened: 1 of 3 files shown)


src/Main.cpp
```diff
@@ -262,20 +266,20 @@ int main(int argc, char** argv) {
...
-                                                   MessageHandlerContainer::getRemoteMsgQueue());
-
-    static CommMonitoringTask s_hostMonitorTask("host_monitor", gc_taskNormalPriority,
-                                                s_hostDispatchTask, s_hostMessageSender,
-                                                BSPContainer::getHostCommInterface);
-    static CommMonitoringTask s_remoteMonitorTask("remote_monitor", gc_taskNormalPriority,
-                                                  s_remoteDispatchTask, s_remoteMessageSender,
-                                                  BSPContainer::getRemoteCommInterface);
+    static MessageSenderTask<HiveMindHostAccumulatorSerializer> s_remoteMessageSender(
+        "remote_send", gc_taskNormalPriority, NULL, MessageHandlerContainer::getRemoteMsgQueue());
```
t-constantin, line 275: What should be not normal priority?

t-constantin (approved):
LGTM, small question but not on the changes themselves

src/Main.cpp
g-lesssard: Tasks of higher priorities like interloc. This "normal" priority is basically the default one that we use for the dispatcher and message senders

src/Main.cpp
```diff
@@ -120,6 +120,7 @@ class MessageDispatcherTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE>
     }
 };
 
+template <typename SerializerType>
```
xgroleau, line 123: Is there a reason for a template instead of injecting a reference of the common interface? By this I mean a `IHiveMindHostSerializer`. It's already created in the comm monitoring, so you could pass this pointer instead of the stream

src/bsp/src/stm32/include/SpiEsp.h
```diff
@@ -47,7 +47,7 @@ class SpiEsp : public ICommInterface {
         uint16_t m_payloadSize;
     } m_inboundMessage, m_outboundMessage;
 
-    std::array<uint8_t, ESP_SPI_MAX_MESSAGE_LENGTH> m_data;
+    std::array<uint8_t, 4 * ESP_SPI_MAX_MESSAGE_LENGTH> m_data;
```
xgroleau, line 50: Why 4 times the message length?

src/Main.cpp
```diff
@@ -144,29 +145,32 @@ class MessageSenderTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
...
+                    if (!messageSender.processAndSerialize()) {
+                        m_logger.log(LogLevel::Warn, "Fail to process/serialize in %s", m_taskName);
+                    }
                 }
             }
         }
     }
 };
 
+template <typename SerializerType>
```
xgroleau, line 166: You could the default template param serializer to a HiveMindHostSerializer, avoiding the need to specify it for the other tasks.

xgroleau (changes requested):
GJ! Minor questions on the modifications

src/bsp/src/stm32/include/SpiEsp.h
g-lesssard: So that we can store 4 messages of the max length while the deserializer reads them

src/Main.cpp
g-lesssard: Not really sure how to do this since the MessageSenderTask is injected in the comm monitoring. I would need to set it with a setter? If so, a reference would not work, it would have to be a pointer...

xgroleau: Yes, as of now we are using a pointer for the stream in the message sender and then create a serializer, so instead we could use a pointer to the interface of serializer.

xgroleau (approved):

PR #99, opened 2021-04-23
Swarinfo 366  UUID in flash

cquesnel:
- UUID is read/saved in flash
- A CMake variable (`-DUUID_OVERRIDE) can be used to set the ID
- If nothing is flashed, ID defaults to 1
- Sector used for user data is not available for the compiler to place code inside


src/bsp/src/stm32/include/PersistedStorageManager.h
```diff
@@ -0,0 +1,31 @@
+#ifndef __PERSISTEDSTORAGEMANAGER_H__
+#define __PERSISTEDSTORAGEMANAGER_H__
+
+#include "PersitedStorage.h"
+#include <logger/ILogger.h>
+
+class PersistedStorageManager {
```
xgroleau, line 7: PersistantStorageManager

src/bsp/src/stm32/src/PersistedStorageManager.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#define DEFAULT_UUID 1
+
+PersistedStorageManager::PersistedStorageManager(ILogger& logger) : m_logger(logger) {}
+
+void PersistedStorageManager::loadFromFlash() {
+    std::memcpy(&m_storage, reinterpret_cast<const void*>(USER_DATA_FLASH_START_ADDRESS),
+                sizeof(m_storage));
+
+    if (UUID_OVERRIDE != UINT16_MAX) {
```
xgroleau, line 16: I think we could use 0 instead of UINT16_MAX for this case since it's already reserved. The user shouldn't set it's id to 0 anyway

src/bsp/src/stm32/hal/src/hal_flash.c
```diff
@@ -0,0 +1,37 @@
...
+}
+
+bool Flash_program(uint32_t address, uint8_t* data, uint32_t size) {
+    if (data == NULL) {
+        return false;
+    }
+
+    HAL_FLASH_Unlock();
+
+    for (unsigned int i = 0; i < size; i++) {
```
xgroleau, line 24: Should we check that the address is in the user data bounds? Or maybe a wrapper that does this check to keep the logic of this class of only writing to flash. (So we could have one wrapper for buzz code and one for settings to make sure both have their separate spaces)

xgroleau (changes requested):
Nice of finally having a flash utils ! Minor comments and questions on checks

src/bsp/src/stm32/src/PersistedStorageManager.cpp
cquesnel: It's because erased flash has a value of 0xFF not 0x0, so this catches a board which has not had an ID set yet

cmake/stm32/linker_ld.cmake
```diff
@@ -16,6 +16,10 @@ _eccmram = .;\n\
 } >CCMRAM AT> FLASH\n\
         ")
 endif()
+
+if(NOT USER_DATA_FLASH_SIZE)
+    set(USER_DATA_FLASH_SIZE "0K")
```
g-lesssard, line 21: Why "OK" and not 0 since you use this in a substraction afterwrards?

src/bsp/src/stm32/src/PersistedStorageManager.cpp
g-lesssard: It makes sense for the read value of `m_storage.m_uuid` but for the UUID_OVERRIDE, since it's a build constant, checking it against 0 would make more sense than UINT16_MAX

g-lesssard (approved):
LGTM, minor comments

src/bsp/src/stm32/src/PersistedStorageManager.cpp
cquesnel: Changed for 0

cmake/stm32/linker_ld.cmake
cquesnel: It's 0K, not OK and it's because the .ld expects a unit after the number

xgroleau (approved):

PR #100, opened 2021-04-28
Feat/swarinfo 420 angle in simulation

cquesnel:
Fix relative azimuth that didn't have the correct sign.
Added the angle between agents that is pushed in the buzz elevation.


xgroleau (approved):
LGTM, though you could use the Math namespace from the BSP, and add the missing functions/constants if needed.

PR #104, opened 2021-05-16
Feat/swarinfo 422 support for h7

xgroleau:


src/bsp/src/stm32/DefaultSettings.cmake
```diff
@@ -7,6 +7,10 @@ if (${COMPILE_STM32_F429ZI})
     set(USER_DATA_FLASH_SIZE "128K")
     set(USER_DATA_FLASH_SECTOR "23")
     set(USER_DATA_FLASH_START_ADDRESS "0x081E0000")
+elseif (${COMPILE_STM32_H735ZG})
+    set(USER_DATA_FLASH_SIZE "128K")
+    set(USER_DATA_FLASH_SECTOR "7")
+    set(USER_DATA_FLASH_START_ADDRESS "0x080E0000")
```
xgroleau, line 13: @cquesnel  Can you double check if that's ok with your confluence?

src/bsp/src/stm32/hal/src/hal_flash.c
```diff
@@ -21,14 +21,14 @@ bool Flash_program(uint32_t address, uint8_t* data, uint32_t size) {
...
+    for (unsigned int i = 0; i < size; i+=4) {
+        if (HAL_FLASH_Program(FLASH_PROGRAM_32_BYTES, address, (uint64_t)(*data)) != HAL_OK) {
             HAL_FLASH_Lock();
             return false;
         }
 
-        address++;
-        data++;
+        address+=4;
+        data+=4;
```
xgroleau, line 31: @cquesnel Fix since we can't program per byte, tagging you so you can double check it.

src/bsp/src/stm32/DefaultSettings.cmake
cquesnel: Seems good

tools/openocd/stm32_h7/stm32_h7.cfg
```diff
@@ -0,0 +1,7 @@
+source [find interface/stlink-v2-1.cfg]
```
cquesnel, line 1: This will not be good as we won't be using a STLink, just keep this in mind

cquesnel (approved):
LGTM

src/bsp/src/stm32/hal/src/hal_flash.c
cquesnel: I don't think it will cause a segfault unless the data we are writing is at the limit of the RAM. If it ever does, we could just pad the structure we are writing so it is a multiple of 4

tools/openocd/stm32_h7/stm32_h7.cfg
xgroleau: True we don't have a nucleo, I'll remove it

PR #107, opened 2021-05-22
SWARINFO-428 BB channel handling

cquesnel:
On bootup, checks which channels are populated and enables them


xgroleau (approved):
LGTM, code is nice and easy to read

PR #109, opened 2021-05-29
Feat/swarinfo 430 UI

cquesnel:
Add driver for IO expander used for buttons and hex display on HB
Add HAL layer to control button callbacks, LEDs, RGB and hex
Add example in BSP where a button press toggles between RGB states

We still need to define what we want to use the UI for and have a high level abstraction that will output everything depending on the state of the firmware.


g-lesssard (approved):
LGTM

src/bsp/src/stm32/src/BSP.cpp
```diff
@@ -1,16 +1,29 @@
...
+}
+
 void BSP::initChip(void* args) {
     (void)args;
 
     Hal_init();
     m_storage.loadFromFlash();
+    dynamic_cast<UserInterface&>(BSPContainer::getUserInterface())
+        .setButtonCallback(Button::BUTTON_0, buttonCallback, this);
+    dynamic_cast<UserInterface&>(BSPContainer::getUserInterface())
```
xgroleau, line 25: Dynamic casts needs the RTTI. What is the overhead of the RTTI? Should we compile with -fno-rtti (for embedded) and just use c style or reinterpret cast when necessary? Valid for all dynamic cast calls

xgroleau (approved):

src/bsp/src/stm32/src/BSP.cpp
cquesnel: Yeah good idead. I added -fno-rtti for C++ targets and changed for a reintrepret_cast. Anyways, this code will not be permanent

PR #112, opened 2021-06-13
Fix/swarinfo 431 distance nan interloc

xgroleau:
Fixes this [issue](https://jira.swarmus.jajservers.duckdns.org/browse/SWARINFO-431). The issue arise when the HB without a robot (connected to the phone) tries to get a distance.

### DIFF
--- a/src/bsp/src/posix/src/InterlocManager.cpp
+++ b/src/bsp/src/posix/src/InterlocManager.cpp
@@ -10,6 +10,13 @@
 #define ROBOT_CENTER_SUFFIX "/base_link"
 #define HIVEBOARD_SUFFIX "/hiveboard"
 
+bool isTfValid(const tf2::Transform& tf){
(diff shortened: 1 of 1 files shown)


src/bsp/src/posix/src/InterlocManager.cpp
```diff
@@ -123,6 +132,14 @@ void InterlocManager::gazeboUpdateCallback(const gazebo_msgs::ModelStates& msg)
 
         tf2::Transform relTransform = currentAgentTf.inverseTimes(distantAgentTf);
 
+        if (!isTfValid(relTransform)) {
```
cquesnel, line 135: So if the transform doesn't exist, we will log at every interloc callback (10 Hz). Could we just do it a few times and then just give up on the interloc for this agent?

cquesnel (approved):

cquesnel (approved):

PR #118, opened 2021-09-02
Fix/hiveboard bringup

cquesnel:
Makes the following things work on the HiveBoard:

- STM32 Debugging
- UI
- Decawaves
- Ethernet

New Bug:
RNG initialization fails (calls error handler)


src/bsp/src/stm32/hal/src/hal.c
```diff
@@ -18,7 +18,7 @@ void Hal_initMcu() {
 
     /* Reset of all peripherals, Initializes the Flash interface and the Systick. */
     HAL_Init();
-    MX_RNG_Init();
+    // MX_RNG_Init();
```
xgroleau, line 21: Uncomment? Or remove if initialized somewhere else

cquesnel: Ah, wait missing a TODO. But it's because the init fails

PR #120, opened 2021-09-09
Fix USB clock source

cquesnel:
Fixes the USB clock source on the HiveBoard


src/bsp/src/stm32/hal/src/hal.c
```diff
@@ -32,6 +32,8 @@ void Hal_initMcu() {
 
     PHal_initMcu();
     UartPrint_init();
+
+    MX_USB_DEVICE_Init();
```
xgroleau, line 36: Where was it initialized before?

xgroleau (approved):
LGTM 👍 mostly generated code anyway

src/bsp/src/stm32/hal/src/hal.c
cquesnel: It seems like I had deleted it at some point

PR #121, opened 2021-09-15
Feat/swarinfo 490 common interface for UI

xgroleau:
Added the code to manage the User Interface. Though the states are not managed by the apps

### DIFF
--- a/src/bsp/src/posix/src/UserInterface.cpp
+++ b/src/bsp/src/posix/src/UserInterface.cpp
@@ -38,7 +38,8 @@ UserInterface::UserInterface(const IBSP& bsp) : m_bsp(bsp), m_mutex(10) {}
 Mutex& UserInterface::getPrintMutex() { return m_mutex; }
 
 void UserInterface::flush() {
-    ROS_INFO("[HM: %d] %s", m_bsp.getUUId(), m_accumulatedString.c_str());
+    ROS_INFO("[HM: %d] %s %s", m_bsp.getUUId(), uiStateToString().c_str(),
+             m_accumulatedString.c_str());
     m_accumulatedString = "";
 }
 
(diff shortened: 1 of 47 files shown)


src/bsp/src/posix/src/UserInterface.cpp
```diff
@@ -53,3 +82,50 @@ int UserInterface::printLine(const char* format, va_list args) {
...
+void UserInterface::setHexDisplay(uint8_t value) { m_uiState.m_hexDisplay = value; }
+
+void UserInterface::setButtonCallback(Button button,
+                                      buttonCallbackFunction_t callback,
+                                      void* context) {
+    ros::NodeHandle nodeHandle = ros::NodeHandle("~");
+    uint buttonId = static_cast<uint>(button);
+
+    std::string buttonTopic =
+        nodeHandle.param("buttonTopic", std::string("/agent_1/user_interface/button"));
```
cquesnel, line 101: Shouldn't this be based on the agent UUID?

cquesnel (approved):

src/bsp/src/posix/src/UserInterface.cpp
xgroleau: I've matched what we done in the HiveMindBridge ros node in SwarmUS ROS [here](
https://github.com/SwarmUS/SwarmUS-ROS/blob/4d047e4b6a86e77aa659400d2cbb6df7cd5941f9/src/hive_mind_bridge/src/main.cpp#L78)

cquesnel: Probably a bug there too.
Because right now you are subscribing to the same topic for every running HM which is problematic

xgroleau: The second argument is just the default argument if the param is not available in the node handle, it's possible to send a different param to each nodehandle. That's what we do in the launch files

PR #127, opened 2021-09-18
Fix

xgroleau:
fix ui not working on buzz


src/application-interface/src/ApplicationInterface.cpp
```diff
@@ -39,8 +39,9 @@ void ApplicationInterface::setSystemConnectionState(ConnectionState state) {
 }
 
 void ApplicationInterface::setSystemDeviceState(DeviceState state) {
-    LockGuard lock(m_mutex);
+    m_mutex.lock();
```
g-lesssard, line 42: Out of curiosity, why was this changed? Is it an issue with the LockGuard from Propolis or did the member really needed to be accessed earlier than the return of the function?

g-lesssard (approved):
LGTM, small question concerning mutex

src/application-interface/src/ApplicationInterface.cpp
g-lesssard: Or before the syscall to setHexDisplay I would guess?

xgroleau: Yea it's because setHexDisplay calls it so there would be a deadlock

PR #129, opened 2021-09-20
Feat/swarinfo 488 vm reset button

xgroleau:
VM resets with the button 1

### DIFF
--- a/src/Main.cpp
+++ b/src/Main.cpp
@@ -81,11 +81,12 @@ class BittyBuzzTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
         std::array<std::reference_wrapper<IBittyBuzzLib>, 3> buzzLibraries{
             {bbzFunctions, mathLib, uiLib}};
 
-        // set state and register btn function
-        m_deviceStateUI.setDeviceState(DeviceState::Ok);
+        // register btn function
         m_buttonCallbackRegister.setCallback(resetVmButtonCallback, &m_resetVm);
 
         while (true) {
+            m_deviceStateUI.setDeviceState(DeviceState::Ok);
+
             if (!m_bittybuzzVm.init(buzzLibraries.data(), buzzLibraries.size())) {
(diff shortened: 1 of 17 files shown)


src/Main.cpp
```diff
@@ -68,39 +82,50 @@ class BittyBuzzTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
...
-                break;
-            }
-            case BBVMRet::VmErr: {
-                m_logger.log(LogLevel::Error, "BBZVM failed to step. state: %s err: %s",
+            m_deviceStateUI.setDeviceState(DeviceState::Ok);
+
+            if (!m_bittybuzzVm.init(buzzLibraries.data(), buzzLibraries.size()) ||
+                !m_bittybuzzVm.start()) {
+
+                m_logger.log(LogLevel::Error, "",
```
cquesnel, line 94: Not logging anything here?

src/Main.cpp
```diff
@@ -68,39 +82,50 @@ class BittyBuzzTask : public AbstractTask<10 * configMINIMAL_STACK_SIZE> {
...
+                }
+                case BBVMRet::VmErr: {
+                    m_logger.log(LogLevel::Error, "BBZVM failed to step. state: %s err: %s",
+                                 BittyBuzzSystem::getStateString(m_bittybuzzVm.getState()),
+                                 BittyBuzzSystem::getErrorString(m_bittybuzzVm.getError()));
+                    m_deviceStateUI.setDeviceState(vmErrorToDeviceState(m_bittybuzzVm.getError()));
+                    break;
+                }
+                }
+                Task::delay(100);
```
cquesnel, line 124: Should this be in a constant somewhere? It seems like something we might want to let the user configure

src/Main.cpp
xgroleau: idk why the format was removed, fixed

src/Main.cpp
xgroleau: It's in the next PR don't worry. And the VM will stop printing endlessy when it cashes

PR #130, opened 2021-09-20
Feat/swarinfo 474 interloc raw data output

cquesnel:
Handle new InterlocMessages (https://github.com/SwarmUS/Propolis/pull/55) and add new states to allow reading/sending angle data and accessing those states via messages


src/bsp/src/stm32/src/interloc/src/Decawave.cpp
```diff
@@ -305,32 +336,32 @@ void Decawave::retrieveRxFrame(UWBRxFrame* frame) {
...
+    //
+    //    // Read the frame into memory without the CRC16 located at the end of the frame
+    //    dwt_readrxdata(frame->m_rxBuffer.data(), m_callbackData.datalength - UWB_CRC_LENGTH, 0);
+    //    getRxTimestamp(&frame->m_rxTimestamp);
+    //
+    //    dwt_readfromdevice(RX_TTCKO_ID, 4, 1, &(frame->m_sfdAngleRegister));
+    //    // Read information needed for phase calculation
+    //    uint16_t firstPathIdx = dwt_read16bitoffsetreg(RX_TIME_ID, RX_TIME_FP_INDEX_OFFSET);
+    //    // Read one extra byte as readaccdata() returns a dummy byte
+    //    dwt_readaccdata(frame->m_firstPathAccumulator, 5, firstPathIdx);
```
xgroleau, line 364: Is this still supposed to be commented out?

xgroleau (approved):

src/bsp/src/stm32/src/interloc/src/Decawave.cpp
cquesnel: Yeah, I'll be moving this to another function in the next PR so I'm leaving it here for reference. There's an equivalent version being done in the ISR atm.

PR #131, opened 2021-09-20
Fix/swarinfo 529 fix vm dump uart buffer

xgroleau:
Added some fixes on the crash dump of the vm and avoided the logger to get flooded with messages when the vm has an error (i.e. when the queue is full) Also the user can add it's own bst files. The delay can be set via the cmake config or ROS params


src/bittybuzz/src/BittyBuzzSystem.cpp
```diff
@@ -10,6 +11,9 @@ IBittyBuzzMessageService* BittyBuzzSystem::g_messageService = NULL;
 IBSP* BittyBuzzSystem::g_bsp = NULL;
 IUserUI* BittyBuzzSystem::g_userUI;
 
+static constexpr uint16_t VM_DUMP_LOG_WAIT_TIME =
+    5; // Wait time between logs for the VM dump to avoid fulling the buffer
```
cquesnel, line 15: Maybe add the unit in the comment or the variable name

src/bsp/src/posix/src/SettingsContainer.cpp
```diff
@@ -49,3 +49,8 @@ LogLevel SettingsContainer::getLogLevel() {
 
     return LogLevel::Info;
 }
+
+uint16_t SettingsContainer::getBBZVMStepDelay() {
+    auto handle = getRosNodeHandle();
+    return handle->param("bbzvm_step_delay", 10);
```
cquesnel, line 55: Add unit for the delay?

cquesnel (approved):

PR #134, opened 2021-09-23
Migrated to delayuntil and fix deadlock

xgroleau:


cmake/propolis/common.cmake
```diff
@@ -13,7 +13,7 @@ function(propolis_fetch_populate)
         ${PROJECT_NAME}_propolis
 
         GIT_REPOSITORY  https://github.com/SwarmUS/Propolis
-        GIT_TAG         a7585a3
+        GIT_TAG         559f1c6
```
cquesnel, line 16: Don't forget to change

cquesnel (approved):
LGTM

PR #139, opened 2021-09-25
Fix/flash programming

cquesnel:
Fix writing to flash for the H7.

- Disable D-Cache as it causes issues
- Change the function call to HAL_FLASH_Program() as it takes a pointer to the data on the H7 instead of the data itself
- Change how many bytes are written to flash as we can write single bytes on the F4, but only DoubleWords on the H7

### DIFF
--- a/src/bsp/src/stm32/hal/src/stm32h735zg/hal_init.c
+++ b/src/bsp/src/stm32/hal/src/stm32h735zg/hal_init.c
@@ -82,8 +82,9 @@ static void Hal_initCache() {
     /* Enable I-Cache */
     SCB_EnableICache();
 
-    /* Do not enable D-Cache as it causes a hardfault when trying to read or write to flash */
-    // SCB_DisableDCache();
+    /* Enable D-Cache */
+    SCB_EnableDCache();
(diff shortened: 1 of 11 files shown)


src/bsp/src/stm32/hal/src/stm32h735zg/hal_init.c
```diff
@@ -82,8 +82,8 @@ static void Hal_initCache() {
     /* Enable I-Cache */
     SCB_EnableICache();
 
-    /* Enable D-Cache */
-    SCB_EnableDCache();
+    /* Do not enable D-Cache as it causes a hardfault when trying to read or write to flash */
+    // SCB_DisableDCache();
```
xgroleau, line 86: Is the disable supposed to be commented? Or it's disabled by default?

xgroleau (approved):

src/bsp/src/stm32/hal/src/stm32h735zg/hal_init.c
cquesnel: I just reenabled the cache

PR #141, opened 2021-09-29
Feat/swarinfo 538 remove data read from isr

cquesnel:
Big refactor of how we handle RX on a decawave.

Instead of retrieving all of the information right away, we have to call a function.


src/bsp/src/stm32/src/interloc/include/interloc/Decawave.h
```diff
@@ -114,46 +114,40 @@ class Decawave {
...
      * @param rxTimeoutUs Timeout in microseconds. 0 to block indefinitely
-     * @return True if successful, false otherwise.
+     * @return An UWBRxStatus if successful, false otherwise.
      */
-    bool transmitAndReceive(uint8_t* buf,
-                            uint16_t length,
-                            uint32_t rxAfterTxTimeUs,
-                            UWBRxFrame& frame,
-                            uint16_t rxTimeoutUs);
+    std::optional<UWBRxStatus> transmitAndReceive(uint8_t* buf,
```
g-lesssard, line 120: Why the optional here? Couldn't the error state be handled within the UWBRXStatus?

g-lesssard (approved):
LGTM, altough some question about the need for an optional return type

PR #142, opened 2021-09-29
Feat/swarinfo 556 freertos cpu usage

xgroleau:


src/bsp/src/stm32/hal/src/hal.c
```diff
@@ -14,6 +14,10 @@
 #include <lwip/apps/lwiperf.h>
 #endif
 
+#ifdef RUNTIME_STATS
+uint32_t HAL_getCPUCounter() { return RUNTIME_STATS_TIMER->Instance->CNT; }
```
cquesnel, line 18: Not sure if this should go into a platform specific file, because the CNT register might not be named the same between architectures

src/bsp/src/stm32/src/BSP.cpp
```diff
@@ -3,6 +3,22 @@
 #include "bsp/BSPContainer.h"
 #include <hal/hal.h>
 #include <logger/LoggerContainer.h>
+#include <Task.h>
+
+#ifdef RUNTIME_STATS
+#include "task.h"
+void usageFunctionCallback(void* ctx) {
+    (void)ctx;
+    char buffer[1000];
```
cquesnel, line 12: should we make a big buffer like this static? Seems like a lot of memory to put on the stack. (Or actually the heap as the task is not created as static)

src/bsp/src/stm32/src/BSP.cpp
```diff
@@ -28,6 +44,11 @@ void BSP::initChip(void* args) {
         .setButtonCallback(Button::BUTTON_0, buttonCallback, this);
```
cquesnel, line 44: We could remove these two lines. I don't think they are needed anymore

cquesnel (changes requested):

src/bsp/src/stm32/src/BSP.cpp
xgroleau: Done hough we will still need to have a stack of at least 1000 since the logger creates a buffer of 1000 when printing

cquesnel (approved):

